PR #12, opened 2020-07-10
Add support to JUnit 5 assertions for TestProject cloud report

gileli121:
This pull request add JUnit 5 extension I developed that make the magic that any usage of JUnit `fail()` method or any usage of `AssertionError` exception will be reflected in TestProject cloud report.

How to use:
You need to add `@ExtendWith({TPAssertionsReporter.class})` to be above the test class.

Example:
```java
@ExtendWith({TPAssertionsReporter.class})
class JUnitTest {
	...
	...
	
    @Test
    @DisplayName("Fail test using fail method")
    void testExample() {
        driver.navigate().to("http://google.com/");
        System.out.println("test example");
        fail("Some failure message");
    }
	
	...
	...
}

```

Now the `fail()` method will fail the test also on the cloud report and not only in JUnit 5 :) 
It will also work with `assertEquals()` and any `assert*` 
In addition, it will also work by throwing the `AssertionFailedError` or `AssertionError` exceptions.

image
image
image


build.gradle
```diff
@@ -52,7 +52,7 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.1'
 
     // Module "junit-jupiter-params" of JUnit 5.
-    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params', version: '5.5.1'
+    implementation group: 'org.junit.jupiter', name: 'junit-jupiter-params', version: '5.5.1'
```
gileli121, line 55: I think that this may be an issue (change `testImplementation ` to `implementation`) when using the SDK from the maven repository in case all the `org.junit.jupiter` is included with this SDK. I don't think that it is something you want.

I had no choice. Otherwise, I could not create the  `TPJunit5Extension`. 
If this is a real problem and the org.junit.jupiter will be included in the SDK, do you know any workaround for this?

src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
```diff
@@ -172,7 +189,19 @@ public void step(final String description,
...
+            boolean isExtensionCall = false;
+            for (StackTraceElement trace : traces) {
+                if (trace.getClassName().startsWith("io.testproject.sdk.utils")) {
+                    isExtensionCall = true;
+                    break;
+                }
+            }
+
+            // Proceed only if the call is not from utils
+            if (!isExtensionCall)
```
gileli121, line 200: I wanted to note and forgot that I really really don't like what I did here.
At the time I did this change, I could not think about a better solution but now I have one.

**Why I did this change**
Because I noticed that if `this.driver.getReportingCommandExecutor().reportTest(traces, false)` is called then it will mess the report. I understand that if this line is called as a result of `TPJunit5Extension` than it must NOT be called.

**Solutions I thought about**
1. The first solution is to create another `step()` method. this solution is rejected because such method will be available inside public API and I don't know how to create another `step()` method and hide it from the API at the same time. it must be a public method.
2. The second solution is to use `AgentClient` directly to report the step. This solution is rejected because then I could not find any way to know if the reports are disabled or not. and inside the `step()` method in the `Reporter` class, this logic is available.

Now I have another solution that is solution 3. This solution not done and here is the solution:
3. Extend class `Reporter` and inside the extended class, add specific `step()` method that will not do `this.driver.getReportingCommandExecutor().reportTest(traces, false);` 
The extended class will be somewhere on `io.testproject.sdk.utils` or inside `internal` package

build.gradle
mstrelex, line 55: We do have to reference JUnit library in order to be able creating an extension.
@vitalybu Does it make sense creating a separate repository only for JUnit extensions that people who are actually using JUnit can add as an optional dependency besides the "core" SDK?

gileli121: You did not ask me but anyway I think that because it is going to be the same solution also for TestNG, I think that the good way to go is to make a separate repository.

Anyway, this separate repository. will have to use internal logic that is the `Reporter` inside the core SDK and need somehow a way to make a call to the `step()` method with a custom configuration that the main API is not exposing.
In this case, the requirement is in this case is to not execute code `this.driver.getReportingCommandExecutor().reportTest(traces, false)` 
Need control on this. I did there very ugly workaround.

In addition, it will also need to access the Reporter instance without the need to driver. 
For this, I added the small change that is lines

```java
    public Reporter(final ReportingDriver driver, final AgentClient agentClient) {
        .....
        .....
        reporter = this;
    }
```

```java
    public static Reporter getInstance() {
        return reporter;
    }

```

src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
gileli121: Update: I did solution 3 which is the best solution I found. See the newer commit.

build.gradle
gileli121: Update:
@mstrelex suggested to replace it to `api` instead of `implementation` and asked me to test it as a local repository and to make sure it works without issues on  JUnit5 & TestNG projects.

It works perfectly without any issues at all.
And even the extension in this PR works fine.

I am not sure I am fully understand what `api` means in this case. 
I assume that it solves the issue of including all JUnit library inside the core SDK.

gileli121: Marat and I found that `implementation` is the right way to do it.

mstrelex:
Please create a new PR with the following:

1. JUnit configuration change in `build.gradle`
2. An interface that is annotated with the Extension and forcing to implement the `getDriver` method to gain access to the Reporter / AgentClient
1. Signed off clean commits
1. A commit for a test that demonstrates the usage (implementing hte interface annotated with the extension)
1. A commit that adds the relevant explanations on how to use the interface + the extension to README.md
1. A commit that adds the relevant changes to the HISTORY.md file

A few points:

* Please name the Extension as `io.testproject.sdk.internal.reporting.extensions.junit.ExceptionsReporter`
* Please name the Interface (that is annotated with the Extension) as `io.testproject.sdk.interfaces.junit.ExceptionsReporter`

PR #15, opened 2020-07-15
Use unnamed entry for project name when there is no package to infer from

mstrelex:
It's is not mandatory to place classes in a package, not under `src > main` neither under `src > test`.

If this is the case, infer logic using `getPackageName()` method on `Class` objects, will get an empty string that will be used for project name when opening a development session with the Agent.

This will result in a missing information for report creation and session will be initialized without reporting responding to SDK with 202 HTTP status. Any further reporting commands will result in error 406.

Added logic to use the _unnamed_ entry (`Unnamed Project`) as project name when there is no package wrapping the class.


mstrelex:
Copied PR description to the commit message.

PR #18, opened 2020-07-16
JUnit 4 examples and various documentation/examples fixes.

mstrelex:
# README.md changes
* Added links to simple flow examples
* Mention the need for a developer token earlier

# Examples
* Added JUnit4 dependency and examples.	
* Added a note about the need to deploy the Demo app, before running the iOS test.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,10 @@ dependencies {
     // Java client for Appium Mobile Webdriver
     api group: 'io.appium', name: 'java-client', version: '7.3.0'
 
-    // Use JUnit Jupiter API for testing.
+    // JUnit4
+    testImplementation group: 'junit', name: 'junit', version: '4.13'
+
+    // JUnit5
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.1'
 
     // Module "junit-jupiter-params" of JUnit 5.
--- a/src/test/java/io/testproject/sdk/tests/examples/frameworks/junit4/ExplicitReportTest.java
+++ b/src/test/java/io/testproject/sdk/tests/examples/frameworks/junit4/ExplicitReportTest.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2020 TestProject LTD. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -48,7 +48,10 @@ dependencies {
     // Java client for Appium Mobile Webdriver
     api group: 'io.appium', name: 'java-client', version: '7.3.0'
 
-    // Use JUnit Jupiter API for testing.
+    // JUnit4
+    testCompile group: 'junit', name: 'junit', version: '4.13'
+
+    // JUnit5
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.1'
```
vitalybu, line 55: What's the difference between `testCompile` and `testImplementation`?

src/test/java/io/testproject/sdk/tests/examples/frameworks/junit4/ExplicitReportTest.java
```diff
@@ -0,0 +1,68 @@
...
+    private static ChromeDriver driver;
+
+    @BeforeClass
+    static void setup() throws InvalidTokenException, AgentConnectException, ObsoleteVersionException, IOException {
+        driver = new ChromeDriver(new ChromeOptions(), "Examples", "JUnit4 Example");
+    }
+
+    @Test
+    void testExample() {
+        AutomationFlows.runFlow(driver);
```
vitalybu, line 49: What this does?

build.gradle
mstrelex: `testCompile` is an old syntax

src/test/java/io/testproject/sdk/tests/examples/frameworks/junit4/ExplicitReportTest.java
mstrelex: Run a simple automation flow (used 38 times among this and other tests).
Do you think it should be "unwrapped"?

build.gradle
vitalybu: Why did you choose to use the old syntax in this case?

mstrelex: I copied it from https://mvnrepository.com/artifact/junit/junit/4.13 that pretend to be the real OSS Maven Central, confusing people and using old syntax. Anyway, it has been changed in the latest push. Thanks for noticing this.

src/test/java/io/testproject/sdk/tests/examples/frameworks/junit4/ExplicitReportTest.java
vitalybu: Nope, but maybe the `runFlow` method should be renamed.
It's part of `AutomationFlows` which suggests that there's more than one flow...
But this can be done in a separate PR.

PR #20, opened 2020-08-04
Add a custom capability 'cloud:URL' and relevant documentation

mstrelex:
If cloud:URL is capability set, Agent (0.63.11+) will attempt to initialize a driver using provided URL of a cloud provider such as BrowserStack or SauceLabs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,23 @@ public ChromeDriver(final URL remoteAddress, final ChromeOptions options)
 
 It can also be set using the `TP_AGENT_URL` environment variable.
 
+## Remote (Cloud) Driver
+
+By default, TestProject Agent communicates with the local Selenium or Appium server. \
+In order to initialize a remote driver for cloud providers such as SauceLabs or BrowserStack, \
+a custom capability `cloud:URL` should be set, for example:
+
+```java
+import io.testproject.sdk.drivers.web.ChromeDriver;
+import io.testproject.sdk.drivers.TestProjectCapabilityType;
+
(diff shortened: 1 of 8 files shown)


vitalybu:
How about adding SL to the unit tests we run as part of the CI?

README.md
```diff
@@ -151,6 +151,23 @@ public ChromeDriver(final URL remoteAddress, final ChromeOptions options)
 
 It can also be set using the `TP_AGENT_URL` environment variable.
 
+## Remote (Cloud) Driver
+
+By default, TestProject Agent is using local Selenium and Appium servers. \
```
vitalybu, line 156:
```suggestion
By default, TestProject Agent communicates with the local Selenium or Appium server. \
```

mstrelex:
> How about adding SL to the unit tests we run as part of the CI?

We can do it in a separate PR if we decide to do so and pass the URL+credentials via an environment variable.

README.md
mstrelex: Fixed.

vitalybu:
> We can do it in a separate PR if we decide to do so and pass the URL+credentials via an environment variable.

Why a separate PR?
IMO tests should be submitted with the functionality...

mstrelex:
> > We can do it in a separate PR if we decide to do so and pass the URL+credentials via an environment variable.
> 
> Why a separate PR?
> IMO tests should be submitted with the functionality...

Added: faee77294247377935dfeb64c173c0370cc833ab

PR #22, opened 2020-08-07
Added ExclusionStrategy to be used when serializing command reports.

### LINKED ISSUE #21: Command reporting fails and thus false-fail the test when trying to serialize Actions class
Having a driver instance in a command object makes the command serialization using GSON to fail with an IllegalArgumentException, causing a false-failure of a the automated test.

gileli121:
This exclusion strategy makes sure to skip serialization of any classes implementing the WebDriver interface.
Since this object is redundant when reporting command parameters and doesn’t pass serialization - we exclude it.
See https://github.com/google/gson/issues/1540 for more details on the serialization problem.

Fixes #21


mstrelex:
Can't run CI on forked PRs for now, closing.

PR #26, opened 2020-08-17
Provide access to AddonsHelper for addon actions execution using a proxy.

mstrelex:


mstrelex:
@vitalybu 

Should we include some explanation in README.md regarding Addons and what are proxies used for?
Perhaps we can take it from https://github.com/testproject-io/addons#addons ?

mstrelex:
@vitalybu It's a draft until we we release code/proxy generators.

vitalybu (approved):

PR #27, opened 2020-08-18
Report screenshot along with failed command

### LINKED ISSUE #19: Add a screenshot to failing command report
SDK should submit a step report with a screenshot when the driver command fails.

mstrelex:
Fixes #19 

This change requires Agent 0.63.13+

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/rest/messages/DriverCommandReport.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/messages/DriverCommandReport.java
@@ -44,6 +44,28 @@ public final class DriverCommandReport extends Report {
     private final boolean passed;
 
     /**
+     * Screenshot as base64 string.
+     */
+    private String screenshot;
+
+    /**
+     * Getter for {@link #screenshot} field.
+     *
+     * @return value of {@link #screenshot} field
+     */
+    public String getScreenshot() {
+        return screenshot;
+    }
+
+    /**
+     * Setter for {@link #screenshot} field.
+     * @param screenshot Screenshot (as base64 string) to be set.
+     */
+    public void setScreenshot(final String screenshot) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/testproject/sdk/internal/rest/messages/DriverCommandReport.java
```diff
@@ -43,6 +43,28 @@
...
+     */
+    public String getScreenshot() {
+        return screenshot;
+    }
+
+    /**
+     * Setter for {@link #screenshot} field.
+     * @param screenshot Screenshot (as base64 string) to be set.
+     */
+    public void setScreenshot(final String screenshot) {
```
vitalybu, line 64: The SDK actually takes the screenshot and passes it to the Agent?
Why not take the screenshot on the Agent's side?

mstrelex: Timing - reports are sent asynchronously and screenshot may not reflect actual UI state.

vitalybu: Makes sense. But screenshots can be large (MBs?), maybe the screenshot payload should be compressed instead of converted to base64?

mstrelex: It can reach several megabytes theoretically, however it's an intranet communication.
This is based on the same mechanism Selenium / Appium use - transfer the screenshot as base64 from server to client, and the same as we do when reporting getScreenshot command that has the screenshot data in it's parameters as base64 string.

vitalybu: Ok

vitalybu (approved):

Issue #25, opened 2020-08-11
Assertions are not reported


mstrelex:
Framework assertions such as `fail()` or `assertEquals()` from JUnit, are not reflected in reports.

mstrelex:
#28 addresses JUnit5, other such as JUnit4 and TesNG should be developed

DavidG8168:
#63 Addressed TestNG.
#97  Addresses JUnit4.

PR #28, opened 2020-08-18
Handle JUnit5 assertions and exceptions, including them in reports as failed steps.

mstrelex:
Addresses #25 

Credits go to https://github.com/gileli121

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/reporting/extensions/junit5/ExceptionsReporter.java
+++ b/src/main/java/io/testproject/sdk/internal/reporting/extensions/junit5/ExceptionsReporter.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2020 TestProject LTD. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package io.testproject.sdk.internal.reporting.extensions.junit5;
+
+import io.testproject.sdk.drivers.ReportingDriver;
+import io.testproject.sdk.internal.rest.messages.StepReport;
+import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.openqa.selenium.WebDriverException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.util.StringUtils;
+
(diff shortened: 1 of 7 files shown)


src/main/java/io/testproject/sdk/internal/reporting/extensions/junit5/ExceptionsReporter.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Report thrown exceptions after test execution.
+     */
+    @Override
+    public void afterTestExecution(final ExtensionContext context) throws Exception {
+
+        Optional<?> obj = context.getTestInstance();
+        if (obj.isEmpty()) {
+            LOG.error("Got an empty object");
```
vitalybu, line 55: What does this mean? The log message is absolutely useless...

src/main/java/io/testproject/sdk/internal/reporting/extensions/junit5/ExceptionsReporter.java
```diff
@@ -0,0 +1,112 @@
...
+        if (context.getExecutionException().get() instanceof WebDriverException) {
+            // WebDriverException is already handled by the reporting executor
+            return;
+        }
+
+        Throwable exception = context.getExecutionException().get();
+        if (exception.getClass().getPackage().getName().startsWith(TESTPROJECT_TLD)) {
+            // Do not report SDK internal exceptions
+            return;
+        }
```
vitalybu, line 90: This code can probably be organised a little differently:

```suggestion
        if (context.getExecutionException().isEmpty()) {
            // Proceed only if an exception was thrown
            return;
        }

        Throwable exception = context.getExecutionException().get();
        
        // Do NOT report WebDriverException as it is already handled by the reporting executor
        // Do NOT report SDK internal exceptions
        if (exception instanceof WebDriverException ||
            exception.getClass().getPackage().getName().startsWith(TESTPROJECT_TLD)) {
            return;
        }
```

src/main/java/io/testproject/sdk/internal/reporting/extensions/junit5/ExceptionsReporter.java
```diff
@@ -0,0 +1,112 @@
...
+
+        // Skip reporting when disabled, just log it.
+        if (driver.getReportingCommandExecutor().isReportsDisabled()) {
+            LOG.trace("Step [{}] - [{}]", resultDescription, false);
+            return;
+        }
+
+        // Finally, submit the report
+        StepReport report = new StepReport(resultDescription, null, false, null);
```
vitalybu, line 107: What about screenshots in this case?

src/main/java/io/testproject/sdk/internal/reporting/extensions/junit5/ExceptionsReporter.java
mstrelex: Agree, it should be something else... not sure though in what scenario it might return null, as the execution is invoked **after** Test execution finishes.

src/main/java/io/testproject/sdk/internal/reporting/extensions/junit5/ExceptionsReporter.java
mstrelex: Screenshots make sense when commands fail (e.g. searching for element that is not there), but for assertions?

vitalybu (approved):

PR #32, opened 2020-08-31
Socket Manager

mstrelex:
SocketManager is a singleton class that will manage TCP socket connection to the Agent.
It will open the socket connection only once and will close it when process exits.
This will allow "session reuse", by restarting the driver only and keeping it open.

Resolves issue https://github.com/testproject-io/python-sdk/issues/25 that was reported for Python as well.

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
@@ -26,6 +26,7 @@ import io.testproject.sdk.internal.reporting.inferrers.GenericInferrer;
 import io.testproject.sdk.internal.reporting.inferrers.InferrerFactory;
 import io.testproject.sdk.internal.rest.messages.*;
 import io.testproject.sdk.internal.rest.serialization.DriverExclusionStrategy;
+import io.testproject.sdk.internal.tcp.SocketManager;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
@@ -39,6 +40,7 @@ import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
+import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.WebDriverException;
@@ -51,7 +53,6 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
-import java.net.Socket;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
@@ -105,6 +106,11 @@ public final class AgentClient implements Closeable {
     private static final int ADDON_EXECUTION_SOCKET_TIMEOUT_MS = 60 * 1000;
 
     /**
+     * Minimum Agent version that support session reuse.
+     */
+    private static final String MIN_SESSION_REUSE_CAPABLE_VERSION = "0.64.20";
+
+    /**
      * Logger instance.
      */
     private static final Logger LOG = LoggerFactory.getLogger(AgentClient.class);
@@ -113,6 +119,11 @@ public final class AgentClient implements Closeable {
      * AgentClient instance used by active driver.
      */
     private static AgentClient instance;
+
+    /**
+     * Class member to store Agent version obtained when session is initialized.
+     */
+    private static String version;
 
     /**
      * An instance of the Google JSON serializer to serialize and deserialize objects.
@@ -140,6 +151,11 @@ public final class AgentClient implements Closeable {
     private final CloseableHttpClient httpClient;
 
     /**
+     * Shutdown thread instance.
+     */
+    private final Thread shutdownThread;
+
+    /**
      * Future to keep the async task of starting the reports queue.
      */
     private Future<?> reportsQueueFuture;
@@ -153,11 +169,6 @@ public final class AgentClient implements Closeable {
      * Class member to store Agent session details.
      */
     private AgentSession session;
-
-    /**
-     * Class member to hold an instance of a TCP socket between the SDK and the Agent.
-     */
-    private Socket socket;
 
     /**
      * Project name to report.
@@ -230,12 +241,35 @@ public final class AgentClient implements Closeable {
 
         // Start reports queue
         if (!disableReports) {
-            this.reportsQueue = new ReportsQueue(this.httpClient);
+            this.reportsQueue = new ReportsQueue(this.httpClient, this.getSession().getSessionId());
             this.reportsQueueFuture = reportsExecutorService.submit(this.reportsQueue);
         }
 
         // Make sure to exit gracefully
-        Runtime.getRuntime().addShutdownHook(new Thread(this::close));
+        shutdownThread = new Thread(this::close);
+        Runtime.getRuntime().addShutdownHook(shutdownThread);
+    }
+
+    /**
+     * Determine whether the Agent support session reuse.
+     * @return True if session can be reused, otherwise False.
+     */
+    private static boolean canReuseSession() {
+        if (version == null) {
+            return false;
+        }
+
+        boolean result = false;
+        try {
+            ComparableVersion agentVersion = new ComparableVersion(version);
+            result = agentVersion.compareTo(new ComparableVersion(MIN_SESSION_REUSE_CAPABLE_VERSION)) >= 0;
+            LOG.trace("Agent [{}] {} session re-use", version, result ? "supports" : "does not support");
+            return result;
+        } catch (IllegalArgumentException e) {
+            LOG.error(e.getMessage(), e);
+        }
+
+        return result;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


mstrelex:
Fails because of this:

`Failed to receive development socket connection in timely manner`

As DevelopmentManager still expects a new TCP socket connection, even when already connected.

src/main/java/io/testproject/sdk/internal/tcp/SocketManager.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public void openSocket(final String host, final int port) throws AgentConnectException {
+        if (socket != null && socket.isConnected()) {
+            LOG.debug("Development socket is already connected.");
+            return;
+        }
+
+        try {
+            LOG.trace("Connecting to Agent socket: {}:{}", host, port);
+            socket = new Socket(host, port);
```
vitalybu, line 103: Is there a timeout for the connection attempt?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -396,22 +396,19 @@ public static AgentClient getClient(final URL remoteAddress, final String token,
...
                     capabilities.getCapability(TP_GUID))) {
 
-
                 // No instance yet or it's for another driver and needs to be re-initialized
+                AgentClient previousInstance = instance;
                 instance = new AgentClient(remoteAddress, token, capabilities, reportSettings, disableReports);
+
+                if (previousInstance != null) {
+                    previousInstance.close();
+                }
```
vitalybu, line 405: Why not close the old instance before creating a new one?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -56,23 +56,32 @@
...
     private boolean running;
 
     /**
      * Initializes a new instance of the class.
+     *
      * @param httpClient HTTP client ot use for communicating with the Agent.
+     * @param sessionId  Driver session ID.
      */
-    public ReportsQueue(final CloseableHttpClient httpClient) {
+    public ReportsQueue(final CloseableHttpClient httpClient, final String sessionId) {
```
vitalybu, line 75: Why do you need the Session ID here?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -169,6 +164,11 @@
      */
     private String jobName;
 
+    /**
+     * Flag to indicate that this instance has been already closed.
+     */
+    private boolean closed;
```
vitalybu, line 170: Why do we care if it's already closed or not?

vitalybu:
> Fails because of this:
> 
> `Failed to receive development socket connection in timely manner`
> 
> As DevelopmentManager still expects a new TCP socket connection, even when already connected.

At what point?

src/main/java/io/testproject/sdk/internal/tcp/SocketManager.java
mstrelex: Added.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Done.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Avoid having both the shutdown thread and the explicit call doing same things multiple times.

mstrelex:
> > Fails because of this:
> > `Failed to receive development socket connection in timely manner`
> > As DevelopmentManager still expects a new TCP socket connection, even when already connected.
> 
> At what point?

When first driver gets closed, and the second is being created.
It has been addressed by checking compatible Agent version.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
vitalybu: Checking if the socket is closed isn't enough?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -230,14 +240,35 @@ private AgentClient(final URL remoteAddress, final String token, final Capabilit
...
 
         // Make sure to exit gracefully
         Runtime.getRuntime().addShutdownHook(new Thread(this::close));
     }
 
+    /**
+     * Determine whether the Agent support session reuse.
+     * @return True if session can be reused, otherwise False.
+     */
+    private static boolean sessionCanBeReused() {
```
vitalybu, line 255:
```suggestion
    private static boolean canReuseSession() {
```

Then it'll look better in an `if` statement ;)
```java
if (canReuseSession()) {
    ... reuse the session ...
}
```

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -513,14 +554,8 @@ private void startSession(final Capabilities capabilities, final ReportSettings
...
-        // Connect to Agent TCP socket
-        try {
-            socket = new Socket(this.remoteAddress.getHost(), agentResponse.getDevSocketPort());
-        } catch (IOException e) {
-            LOG.error("Failed connecting to Agent socket at {}:{}",
-                    this.remoteAddress.getHost(), agentResponse.getDevSocketPort(), e);
-            throw new AgentConnectException("Failed connecting to Agent socket", e);
-        }
+        // Open TCP socket
+        SocketManager.getInstance().openSocket(this.remoteAddress.getHost(), agentResponse.getDevSocketPort());
```
vitalybu, line 558: Using WebSockets for this is definitely the right way to go...
Should consider the switch at some point

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -609,6 +644,36 @@ private void handleSessionStartFailure(final CloseableHttpResponse response)
...
+            throw new IllegalArgumentException("Both versions must not be null");
+        }
+        int comparisonResult = 0;
+
+        String[] version1Splits = version1.split("\\.");
+        String[] version2Splits = version2.split("\\.");
+        int maxLengthOfVersionSplits = Math.max(version1Splits.length, version2Splits.length);
+
+        for (int i = 0; i < maxLengthOfVersionSplits; i++) {
+            Integer v1 = i < version1Splits.length ? Integer.parseInt(version1Splits[i]) : 0;
```
vitalybu, line 666: Won't `parseInt()` throw an exception when parsing DEV tag (the one with the git hash)?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -609,6 +644,36 @@ private void handleSessionStartFailure(final CloseableHttpResponse response)
         }
     }
 
+    /**
+     *
+     * Compare two versions.
+     * @param version1 Version to compare.
+     * @param version2 Version to compare to.
+     * @return Negative if first version is lower than second, otherwise positive.
```
vitalybu, line 652: Or 0 if the versions match

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
mstrelex: For better logging.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: No, as it takes a bit more time to close it before the shutdown thread is called and then happens twice.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Changed.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Agree.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Changed to https://www.baeldung.com/java-comparing-versions#comparableVersion

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Irrelevant.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
vitalybu: Critical section?

mstrelex: @vitalybu FYI

vitalybu: FMI what?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -230,12 +247,46 @@ private AgentClient(final URL remoteAddress, final String token, final Capabilit
...
+    /**
+     * Creates a thread to be executed upon shutdown.
+     * @return Shutdown thread.
+     */
+    private Thread createShutdownThread() {
+        return new Thread(() -> {
+            shuttingDown = true;
+            this.close();
+        });
+    }
```
vitalybu, line 268: Why is this a function if it's used only once?

vitalybu (approved):

PR #35, opened 2020-09-23
Generic Driver

mstrelex:
Add support for a non-UI tests using a `GenericDriver`


src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -465,6 +466,78 @@ public static AgentClient getClient(final URL remoteAddress, final String token,
...
+        String responseBody;
+        try {
+            responseBody = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8.name());
+        } catch (IOException e) {
+            LOG.error("Failed reading Agent status response", e);
+            throw new AgentConnectException("Failed to get Agent status", e);
+        }
+
+        // Parse response to an object
+        AgentStatusResponse status = GSON.fromJson(responseBody, AgentStatusResponse.class);
```
vitalybu, line 535: What about `JsonSyntaxException`?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -624,18 +624,24 @@ private void startSession(final Capabilities capabilities, final ReportSettings
...
         // Process Response
         try {
-            MutableCapabilities mutableCapabilities =
-                    new MutableCapabilities(agentResponse.getCapabilities());
+            MutableCapabilities mutableCapabilities;
+            if (agentResponse.getCapabilities() != null) {
+                mutableCapabilities = new MutableCapabilities(agentResponse.getCapabilities());
+            } else {
+                mutableCapabilities = new MutableCapabilities();
+            }
```
vitalybu, line 632: Should this change really be part of this PR?
When does this happen anyway?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -624,18 +624,24 @@ private void startSession(final Capabilities capabilities, final ReportSettings
...
 
             // There's a chance that driver ignored our custom tracking capability
             // And now does't return it with the actual driver capabilities
             // It has to be set again to avoid initializing an new AgentClient instance for these capabilities again
             mutableCapabilities.setCapability(TP_GUID, guid);
-            this.version = agentResponse.getVersion();
+            version = agentResponse.getVersion();
```
vitalybu, line 638: I don't see this being used anyway...

src/main/java/io/testproject/sdk/drivers/GenericDriver.java
```diff
@@ -0,0 +1,312 @@
...
+        DesiredCapabilities capabilities = new DesiredCapabilities();
+        capabilities.setCapability(CapabilityType.PLATFORM_NAME, Platform.ANY);
+
+        String version = AgentClient.getVersion(remoteAddress);
+        if (new ComparableVersion(version).compareTo(
+                new ComparableVersion(MIN_GENERIC_DRIVER_SUPPORTED_VERSION)) < 0) {
+            StringBuilder message = new StringBuilder()
+                    .append("Target Agent version").append(" [").append(version).append("] ")
+                    .append("doesn't support Generic driver. ")
+                    .append("Upgrade to the latest version available and try again.");
```
vitalybu, line 267:
```suggestion
                    .append("Upgrade the Agent to the latest version and try again.");
```

src/main/java/io/testproject/sdk/drivers/GenericDriver.java
```diff
@@ -0,0 +1,312 @@
...
+        // Nothing here...
+    }
+
+    /**
+     * Returns a local command executor.
+     */
+    @Override
+    public ReportingCommandsExecutor getReportingCommandExecutor() {
+        return reportingCommandExecutor;
+    }
```
vitalybu, line 301: Why do we need a `ReportingCommandsExecutor` for a generic driver?
Ca't we simply return `null` here and handle this in the upper layers by checking for `null` and not reporting anything?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: It's a RuntimeException and therefore IntelliJ is not showing a warning :(
Added code that handles it.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Yes, capabilities are absent if the Driver is a GenericDriver (none passed and none returned by the agent).

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: It is used here - 

```
private static boolean canReuseSession() {
        if (version == null) {
            return false;
        }
```

I just removed the `this.` part in this PR, as it's a static variable and not instance based.

src/main/java/io/testproject/sdk/drivers/GenericDriver.java
mstrelex: There a re a lot of methods that use `getReportingCommandExecutor` and define behaviour such as disabling tests reporting, completely disabling reporting, etc... it's much easier to reuse the same mechanism rather than reinventing the wheel and treating GenericDriver differently all over the code.

vitalybu: Not sure I follow. Most of the uses of `getReportingCommandExecutor` are part of the various drivers, which is fine, but outside the drivers I saw about 3-4 calls.
Handling `null` values in these cases makes more sense to me then creating a dummy `ReportingCommandsExecutor` class.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
vitalybu: According to the Java naming conventions, shouldn't there some difference between regular and static variables?
Just from reading the code, it's not clear that this variable is static..

mstrelex: According to the checkstyle rules only when static & final it should be capitalized, otherwise, no. \
Compiler warns when using a static member with `this.` prefix though.

src/main/java/io/testproject/sdk/drivers/GenericDriver.java
mstrelex: Indeed there are many internal usages inside the drivers to set reporting status,
but also there are some places where this status is queries via the `getReportingCommandExecutor()` method:

# ExceptionsReporter 

- `driver.getReportingCommandExecutor().isReportsDisabled()`

# Reporter 

- this.driver.getReportingCommandExecutor().setReportsDisabled(disable)
- this.driver.getReportingCommandExecutor().setCommandReportsDisabled(disable)
- this.driver.getReportingCommandExecutor().setTestAutoReportsDisabled(disable);
- this.driver.getReportingCommandExecutor().setRedactionDisabled(disable);
- this.driver.getReportingCommandExecutor().reportTest(traces, false);

If we remove the "dummy" `ReportingCommandsExecutor`, in all these places, I will have to add specific logic, in case the Driver is a GenericDriver to returns the state.

vitalybu (approved):

PR #44, opened 2020-10-21
Create fake driver service for IE to avoid searching for driver executable.

DavidG8168:
To avoid the webdriver.ie.driver property not set error when constructing the IE driver,
create a dummy service for IE and use it when constructing the driver.

### DIFF
--- a/src/main/java/io/testproject/sdk/drivers/web/InternetExplorerDriver.java
+++ b/src/main/java/io/testproject/sdk/drivers/web/InternetExplorerDriver.java
@@ -17,21 +17,29 @@
 
 package io.testproject.sdk.drivers.web;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.testproject.sdk.drivers.ReportingDriver;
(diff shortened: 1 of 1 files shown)


src/main/java/io/testproject/sdk/drivers/web/InternetExplorerDriver.java
```diff
@@ -447,8 +447,8 @@ private static InternetExplorerDriverService fakeDriverService() {
                     .getDeclaredConstructor(File.class, int.class, ImmutableList.class, ImmutableMap.class);
             serviceConstructor.setAccessible(true);
             return serviceConstructor.newInstance(new File(""), 0, null, null);
-        } catch (NoSuchMethodException | InstantiationException | IllegalAccessException |
-                InvocationTargetException e) {
+        } catch (NoSuchMethodException | InstantiationException | IllegalAccessException
+                | InvocationTargetException e) {
             e.printStackTrace();
```
mstrelex, line 452: Replace with logger

mstrelex (changes requested):

src/main/java/io/testproject/sdk/drivers/web/InternetExplorerDriver.java
DavidG8168: Changed to logger.

mstrelex (approved):

PR #46, opened 2020-10-25
Translates errors in initial session request to an informative error

s-glatshtein:
When a driver is created, a request is made to open a new session. This request can fail completely without returning a status code. The failure can happen because of several different error, and each indicates of a different problem.
We want users to be able to rectify the problem by themselves. To achieve this we should make the returned error more informative to the user, and reduce its clutter

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
@@ -52,9 +52,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.net.HttpURLConnection;
-import java.net.MalformedURLException;
-import java.net.URL;
+import java.net.*;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.concurrent.*;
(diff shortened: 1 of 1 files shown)


src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -658,6 +657,23 @@ private void startSession(final Capabilities capabilities, final ReportSettings
         SocketManager.getInstance().openSocket(this.remoteAddress.getHost(), agentResponse.getDevSocketPort());
     }
 
+    /**
+     * Translates an IOException to an informative exception.
+     * @param e the original exception
+     * @throws AgentConnectException an informative exception
+     */
+    private void translateAgentConnectFailure(final IOException e) throws AgentConnectException {
```
mstrelex, line 665: It would be better if this method will return an `AgentConnectException`

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -597,7 +595,8 @@ private void startSession(final Capabilities capabilities, final ReportSettings
         try {
             response = httpClient.execute(httpPost);
         } catch (IOException e) {
-            throw new AgentConnectException("Failed communicating with the Agent at " + this.remoteAddress, e);
+            translateAgentConnectFailure(e);
+            return;
```
mstrelex, line 599: It's not required, as method above always throws exception.
Instead it would be better if the method will return an exception that will be thrown.

mstrelex (changes requested):

mstrelex (approved):

PR #49, opened 2020-11-25
Convert strings into the relevant type when setting addon action output.

DavidG8168:
Convert the string to the relevant type to avoid error where output field could not be set due to type mismatch.

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/addons/AddonsHelper.java
+++ b/src/main/java/io/testproject/sdk/internal/addons/AddonsHelper.java
@@ -140,7 +140,7 @@ public class AddonsHelper {
             // Use reflection to set the output value
             proxyField.get().setAccessible(true);
             try {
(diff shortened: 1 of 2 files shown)


mstrelex (approved):

src/main/java/io/testproject/sdk/internal/addons/AddonsHelper.java
```diff
@@ -149,4 +149,35 @@ public ActionProxy execute(final ActionProxy action, final By by, final int time
...
         return action;
     }
+
+    /**
+     * Convert string to specified type.
+     * @param clazz target type for conversion..
+     * @param value the value that will be converted.
+     * @return value converted to correct type.
+     */
+    private Object convertToType(final Class<?> clazz, final String value) {
```
mstrelex, line 159: @DavidG8168 Let's handle `NumberFormatException` exception as well.

PR #52, opened 2020-12-02
Handle NumberFormatException when parsing addon execution results.

DavidG8168:
Handle parsing error, log the error and throw an exception to stop the execution.

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/addons/AddonsHelper.java
+++ b/src/main/java/io/testproject/sdk/internal/addons/AddonsHelper.java
@@ -160,24 +160,29 @@ public class AddonsHelper {
         if (Boolean.class == clazz || boolean.class == clazz) {
             return Boolean.parseBoolean(value);
         }
-        if (Byte.class == clazz || byte.class == clazz) {
-            return Byte.parseByte(value);
+        try {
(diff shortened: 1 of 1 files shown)


src/main/java/io/testproject/sdk/internal/addons/AddonsHelper.java
```diff
@@ -157,27 +157,32 @@ public ActionProxy execute(final ActionProxy action, final By by, final int time
...
-            return Long.parseLong(value);
-        }
-        if (Float.class == clazz || float.class == clazz) {
-            return Float.parseFloat(value);
-        }
-        if (Double.class == clazz || double.class == clazz) {
-            return Double.parseDouble(value);
+        try {
+            if (Boolean.class == clazz || boolean.class == clazz) {
+                return Boolean.parseBoolean(value);
```
mstrelex, line 162: This may throw NumberFormatException ?

DavidG8168: @mstrelex It's the only one that doesn't throw any kind of exception, changed the try block.

mstrelex (approved):

PR #58, opened 2020-12-14
Do not close the Selenium session if TP_KEEP_DRIVER_SESSION environment variable is set to true

DavidG8168:
Used to keep the driver session open if running SDK tests through the TestProject recorder.

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/helpers/CustomHttpCommandExecutor.java
+++ b/src/main/java/io/testproject/sdk/internal/helpers/CustomHttpCommandExecutor.java
@@ -20,6 +20,7 @@ package io.testproject.sdk.internal.helpers;
 import io.testproject.sdk.internal.rest.AgentClient;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.remote.Command;
+import org.openqa.selenium.remote.DriverCommand;
 import org.openqa.selenium.remote.HttpCommandExecutor;
 import org.openqa.selenium.remote.Response;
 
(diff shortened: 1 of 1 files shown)


mstrelex:
@DavidG8168 Can't we override the quit() method in the executer instead of changing each driver?

DavidG8168:
@mstrelex Are you referring to the consideration to the quit command in the ReportingCommandsExecutor? 
In there, the quit command closes the AgentClient, which does not close the Selenium Session, but the calls the reporting blocking queue to finish. 
Session closure is called in each driver with super.quit().

src/main/java/io/testproject/sdk/internal/helpers/CustomHttpCommandExecutor.java
```diff
@@ -107,11 +115,18 @@ public Response execute(final Command command) throws WebDriverException {
...
-        try {
-            response = super.execute(command);
-        } catch (IOException e) {
-            throw new WebDriverException(e);
+        boolean isQuitCommand = command.getName().equals(DriverCommand.QUIT);
+        boolean keepSessionOpen = Boolean.parseBoolean(System.getenv(tpKeepDriverSession));
+        Response response = new Response();
+        // Do not close the Selenium session if the incoming command is quit and the TP_KEEP_DRIVER_SESSION
+        // environment variable is set to true.
+        // Used when running tests through the recorder to avoid closing the session.
```
mstrelex, line 123:
```suggestion
        // Used when running tests through the platform to avoid closing the session.
```

mstrelex (approved):

PR #63, opened 2020-12-28
Add support for TestNG assertion reporting.

DavidG8168:
Added support for TestNG reporting and updated documentation.
Closes TP-13815.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,8 @@ Examples of explicit Project & Job names configuration:
 
 Reporting extensions extend the TestProject SDK reporting capabilities by intercepting unit testing framework assertions or exceptions thrown, and reporting them as failing steps.
 
+### JUnit
+
 In order to integrate it, a relevant interface needs to be implemented by your test class, for example:
 
 ```java
@@ -260,6 +262,26 @@ ReportingDriver getDriver();
 
 This method's code, should provide access to the driver that can be used for the reporting. \
 Here is a complete example for a [JUnit 5](/src/test/java/io/testproject/sdk/tests/examples/frameworks/junit5/ExceptionsReportTest.java) based test.
+
+### TestNG
+
+In order to integrate it, your test class will need to implement the relevant interface, in addition you will need to 
+annotate your class with the TestNG @Listener annotation and direct it to the reporter from the SDK, for example:
+
+```java
+import io.testproject.sdk.interfaces.testng.ExceptionsReporter;
+
+@Listeners(io.testproject.sdk.internal.reporting.extensions.testng.ExceptionsReporter.class)
+public class ExplicitReportTest implements ExceptionsReporter
+```
+
+Implementing this interface requires the class to implement the following method:
+
+```java
+ReportingDriver getDriver();
+```
+
+The method should provide access to the driver that is used for reporting.
 
 ## Tests Reports
 
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,7 @@ dependencies {
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params', version: '5.5.1'
 
     // TestNG Testing framework
-    testImplementation group: 'org.testng', name: 'testng', version: '7.1.0'
+    implementation group: 'org.testng', name: 'testng', version: '7.1.0'
 
     // Logback classic module
     testImplementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
--- a/src/main/java/io/testproject/sdk/interfaces/testng/ExceptionsReporter.java
+++ b/src/main/java/io/testproject/sdk/interfaces/testng/ExceptionsReporter.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2021 TestProject LTD. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.testproject.sdk.interfaces.testng;
+
+import io.testproject.sdk.drivers.ReportingDriver;
+
+/**
+ * Implementing this interface enables automatic TestNG assertions reporting.
+ */
+public interface ExceptionsReporter {
+
+    /**
+     * Getter for a reporting driver instance.
+     * @return ReportingDriver instance.
+     */
+    ReportingDriver getDriver();
+}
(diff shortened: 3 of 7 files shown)


src/main/java/io/testproject/sdk/internal/reporting/extensions/testng/TestNGExceptionReporter.java
```diff
@@ -0,0 +1,112 @@
...
+     * Reporting driver.
+     */
+    private static ReportingDriver driver;
+
+    /**
+     * Report thrown exceptions after test execution.
+     * @param result TestNG test result.
+     */
+    @Override
+    public void onTestFailure(final ITestResult result) {
```
mstrelex, line 52: onTestPassed exists?

src/main/java/io/testproject/sdk/internal/reporting/extensions/testng/TestNGExceptionReporter.java
```diff
@@ -0,0 +1,112 @@
...
+        if (!driver.getReportingCommandExecutor().getAgentClient().reportStep(report)) {
+            LOG.error("Failed reporting exception: [{}]", report);
+        }
+    }
+
+    /**
+     * Retrieve the reporting driver from the calling test using reflection.
+     * @param result TestNG test result.
+     */
+    private void getDriver(final ITestResult result) {
```
mstrelex, line 88: Check how Junit assertion reporting is implemented using an interface with getDriver() method

src/main/java/io/testproject/sdk/internal/reporting/extensions/testng/TestNGExceptionReporter.java
DavidG8168: @mstrelex It exists, but the most I would be able to do with it is report: "Test Method 'method_name' passed with no assertion errors".

If that is acceptable, please provide phrasing to the report, if not would be glad to hear another suggestion.

src/main/java/io/testproject/sdk/internal/reporting/extensions/testng/TestNGExceptionReporter.java
DavidG8168: @mstrelex I removed the reflection in favor of implementing an interface to pass the driver to the listener, however unlike JUnit, I have not seen any way to extend the Listener class, and the use of the Listener annotation from TestNG which will point to the listener from the SDK will still be required for the test class.
See README for details.

mstrelex:
@vitalybu please share your opinion - Since `@Listeners` annotation can't be inherited, user will have to do both - add the annotation and also implement the interface.

src/main/java/io/testproject/sdk/internal/reporting/extensions/testng/ExceptionReporter.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Retrieve the reporting driver from the calling test.
+     * @param result TestNG test result.
+     *
+     * @return The reporting driver.
+     */
+    private ReportingDriver getDriver(final ITestResult result) {
+        return ((io.testproject.sdk.interfaces.testng.TestNGExceptionReporter) result.getInstance()).getDriver();
```
mstrelex, line 81: You must first make sure that the class indeed implements this interface and has this method before casting, as the annotation can be placed on any class.

README.md
```diff
@@ -261,6 +263,27 @@ ReportingDriver getDriver();
...
+
+@Listeners(ExceptionReporter.class)
+public class ExplicitReportTest implements TestNGExceptionReporter
+```
+
+Implementing this interface requires the class to implement the following method:
+
+```java
+ReportingDriver getDriver();
+```
```
vitalybu, line 283: Do we have an example for that? Maybe in a unit-test?

vitalybu:
> @vitalybu please share your opinion - Since `@Listeners` annotation can't be inherited, user will have to do both - add the annotation and also implement the interface.

Not sure I understand what you mean here

src/main/java/io/testproject/sdk/internal/reporting/extensions/testng/ExceptionReporter.java
DavidG8168: Agreed, will make the changes.

build.gradle
```diff
@@ -64,7 +64,7 @@ dependencies {
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params', version: '5.5.1'
 
     // TestNG Testing framework
-    testImplementation group: 'org.testng', name: 'testng', version: '7.1.0'
+    implementation group: 'org.testng', name: 'testng', version: '7.1.0'
```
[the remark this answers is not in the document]
DavidG8168: We're already doing this with JUnit, no? 
It's a pretty small addition, I think this should be contained in the SDK instead of splitting into something else.

@mstrelex  what do you think?

README.md
DavidG8168: Will add another example test.

DavidG8168:
@vitalybu What @mstrelex meant was what is your opinion on expecting users to use the Listeners annotation in addition to implementing the interface. 
In JUnit it can be avoided by wrapping the extension with another one, but that can't be done here, if you have an idea on anything we can try to save users needing to use the annotation.

README.md
DavidG8168: Added additional unit test.

build.gradle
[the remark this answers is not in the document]
mstrelex: Agree, we did the same with JUnit, so far no complains for versions conflicts.

vitalybu:
> @vitalybu What @mstrelex meant was what is your opinion on expecting users to use the Listeners annotation in addition to implementing the interface.
> In JUnit it can be avoided by wrapping the extension with another one, but that can't be done here, if you have an idea on anything we can try to save users needing to use the annotation.

I guess it's fine, if we can't find a way around it.
@mstrelex, why did we create the interface in the first place? A test can have multiple drivers?
We can't extract the driver instance automatically in runtime?

mstrelex:
Test can have multiple or even not have a driver field at all and get it from some other class.

DavidG8168:
@vitalybu Can this be merged?

src/main/java/io/testproject/sdk/internal/reporting/extensions/testng/ExceptionReporter.java
vitalybu: Did you?

DavidG8168: @vitalybu Yes, I am checking if the interface is implemented.

```
        if (!(io.testproject.sdk.interfaces.testng.ExceptionsReporter.class.isAssignableFrom(result.getInstance()
                .getClass()))) {
            LOG.error("Class is not implementing the ExceptionReporter interface");
            return;
        }
```

Also I moved the code out of the method, a couple of lines below the snippet above.

```
        ReportingDriver driver = ((io.testproject.sdk.interfaces.testng.ExceptionsReporter) result.getInstance())
                .getDriver();
```

vitalybu: So how come I still see the old commit?
Did you do it in a later commit instead of a fixup to this one?

DavidG8168: @vitalybu Source tree didn't like it as a fixup, I made some changes, now it should be in the same commit:
https://github.com/testproject-io/java-sdk/pull/63/commits/45af91b188a8030f19075c5facb2e7dfcd0602f4

vitalybu (approved):

PR #66, opened 2021-01-03
Close TCP socket connection after reporting queue finishes

DavidG8168:
Close the socket only after finishing reporting to prevent missing reporting driver commands and tests, as well as to prevent a 417 error that is caused by the agent leaving development mode before reporting is finished.

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
@@ -868,6 +868,8 @@ public final class AgentClient implements Closeable {
             reportsExecutorService.shutdown();
         }
 
+        // Make sure to close the socket when exiting.
+        SocketManager.getInstance().closeSocket();
         LOG.info("Session [{}] closed", this.getSession().getSessionId());
     }
 
--- a/src/main/java/io/testproject/sdk/internal/tcp/SocketManager.java
+++ b/src/main/java/io/testproject/sdk/internal/tcp/SocketManager.java
@@ -58,8 +58,6 @@ public final class SocketManager {
      * Private constructor to prevent creating more than one instance.
      */
     private SocketManager() {
-        // Make sure to close the socket when process exits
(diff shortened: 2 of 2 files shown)


src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -868,6 +868,8 @@ public void close() {
             reportsExecutorService.shutdown();
         }
 
+        // Make sure to close the socket when exiting.
+        SocketManager.getInstance().closeSocket();
```
mstrelex, line 872: This will break session reuse.

DavidG8168: @mstrelex I checked the scenario we talked about with:

```
class FirefoxDriverTest {

    /**
     * Driver instance.
     */
    private ChromeDriver driver;

    @BeforeEach
     void setup() throws InvalidTokenException, AgentConnectException, ObsoleteVersionException, IOException {
        driver = new ChromeDriver("TOKEN", new ChromeOptions(), "Project");
    }

    @Test
    void testExample() {
        AutomationFlows.runFlow(driver);
        driver.report().step("Test 1");
    }

    @Test
    void testExample2() {
        AutomationFlows.runFlow(driver);
        driver.report().step("Test 2");
    }
}
```


Both before and after the change, the tests have two different reports:

image

First pair are with changes second pair without.

mstrelex: Both SDK and Agent versions should be > 0.64, did you set TP_DEBUG_SDK_VERSION ?

DavidG8168: @mstrelex Yes,

`2021-01-05 10:55:22.011 DEBUG SessionRequest       Using value [0.65] from TP_DEBUG_SDK_VERSION environment variable
`

PR #70, opened 2021-01-13
Test name is not reported when driver.quit() isn't called after test execution

DavidG8168:


src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public void addShutdownThread(final int priority, final Runnable shutdownThread) {
+        shutdownThreads.put(priority, shutdownThread);
+    }
+
+    /**
+     * Remote the shutdown thread from the manager.
+     * @param priority of the thread to be removed.
+     */
+    public void remoteShutdownThread(final int priority) {
```
mstrelex, line 60: Remote?

mstrelex (changes requested):

mstrelex:
@vitalybu Please have a look

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
DavidG8168: Typo, fixed.

src/main/java/io/testproject/sdk/internal/helpers/DriverShutdownThread.java
```diff
@@ -24,7 +24,7 @@
 /**
  * A thread used to handle a graceful shutdown and driver closure.
  */
-public class DriverShutdownThread extends Thread {
+public class DriverShutdownThread {
```
vitalybu, line 27: This class kinda feels unnecessary now.
Each driver instance should simply register a lambda function with the `ShutdownThreadManager` and close itself...
Maybe use the self object hash as a key to support unregistering.

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
```diff
@@ -0,0 +1,101 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.TreeMap;
+
+/**
+ * Will be used to manage the driver and AgentClient shutdown threads to ensure
+ * the AgentClient will not close before the driver finishes reporting.
+ */
+public final class ShutdownThreadManager extends Thread {
```
vitalybu, line 30: I don't really think we need a numeric priority here, as we only have two so called "priorities".
We want to make sure that all the drivers close before shutting down the `AgentClient`.
So maybe it's better to have a simple list of drivers and an instance of the `AgentClient`.
Then we can simply close the drivers first without any arbitrary number priorities.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -151,7 +152,7 @@
     /**
      * Shutdown thread instance.
      */
-    private final Thread shutdownThread;
+    private final Runnable shutdownThread;
```
vitalybu, line 155: This can removed (as it's a simply lambda: `() -> close(true);`)

src/main/java/io/testproject/sdk/internal/helpers/DriverShutdownThread.java
DavidG8168: Removed it, everything is done now through the manager.
Drivers threads are registered with their hash.

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
DavidG8168: Drivers have been separated from the client, client will be stored separately.
After the drivers are shutdown, the client will be shutdown.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
DavidG8168: Removed.

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
```diff
@@ -0,0 +1,128 @@
...
+     * Logger instance.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(ShutdownThreadManager.class);
+
+    /**
+     * Add Agent Client shutdown thread to the manager.
+     *
+     * @param shutdownThread of the agent client.
+     */
+    public void addClientShutdownThread(final Runnable shutdownThread) {
```
vitalybu, line 57: The class is called `ShutdownThreadManager`, so no reason adding `...ShutdownThread()` to all the methods.

```suggestion
    public void addAgentClient(final Runnable shutdownThread) {
```

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
```diff
@@ -0,0 +1,128 @@
...
+        agentClientShutdownThread = shutdownThread;
+    }
+
+    /**
+     * Remove Agent Client shutdown thread from the manager.
+     */
+    public void removeClientShutdownThread() {
```
vitalybu, line 64:
```suggestion
    public void removeAgentClient() {
```

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Add a driver shutdown thread to the manager.
+     *
+     * @param hash                 of the driver thread.
+     * @param driverShutdownThread the shutdown thread.
+     */
+    public void addDriverShutdownThread(final int hash, final Runnable driverShutdownThread) {
+        driverShutdownThreads.put(hash, driverShutdownThread);
+    }
```
vitalybu, line 76: By passing an `Object`, you can make the calling code more readable, as it'll look like:
```java
X.addDriver(this, this::stop);
```
Instead of:
```java
X.addDriver(this.hashCode(), this::stop);
```

The ShutdownManager is the one which decided to use `hashCode()` now, but in the future it may decide to use a different key...

```suggestion
    public void addDriver(final Object driver, final Runnable shutdownMethod) {
        driverShutdownThreads.put(driver.hashCode(), shutdownMethod);
    }
```

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
DavidG8168: Makes sense, addressed.

src/main/java/io/testproject/sdk/internal/helpers/ShutdownThreadManager.java
DavidG8168: Changed.

vitalybu (approved):

PR #74, opened 2021-01-27
Fix Issue Where Auto Reports Were Not Disabled When Running Cucumber

DavidG8168:
getBoolean always performs System.getProperty internally, while using it externally, it would check
for a system property called "true"/"false" internally and always return false since it does not exist,
causing the auto reports to never be disabled.

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
+++ b/src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
@@ -58,7 +58,7 @@ public final class Reporter {
     public Reporter(final ReportingDriver driver, final AgentClient agentClient) {
         this.agentClient = agentClient;
         this.driver = driver;
-        if (Boolean.getBoolean(System.getProperty("TP_DISABLE_AUTO_REPORTS"))) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
```diff
@@ -58,7 +58,7 @@
     public Reporter(final ReportingDriver driver, final AgentClient agentClient) {
         this.agentClient = agentClient;
         this.driver = driver;
-        if (Boolean.getBoolean(System.getProperty("TP_DISABLE_AUTO_REPORTS"))) {
+        if (Boolean.getBoolean("TP_DISABLE_AUTO_REPORTS")) {
```
vitalybu, line 61: Is it case sensitive? What if the value of the property is `TRUE` or `True`?

DavidG8168: It is case sensitive, True, TRUE, true, TrUe -> is converted to true boolean.
Everything else including null -> false.

mstrelex (approved):

PR #77, opened 2021-01-27
docs: added explanation on how to package unit tests.

mstrelex:


README.md
```diff
@@ -550,6 +550,23 @@ in your program's Run/Debug configuration as :
...
+  
+# Package & Upload Tests to TestProject
+
+Tests can be executed locally using the SDK and remotely from the platform. \
+Before uploading your Tests, they should be packaged into a JAR.
+
+This JAR must contain all the dependencies (including TestProject SDK) and your unit tests (JUnit / TestNG). \
+Since unit Tests are not packaged by default, they must be included explicitly during build.
+
+Here's an example of how this can be done with Gradle in build.gradle:
```
vitalybu, line 562: "this" needs a better explanation.
What exactly those two lines do...

vitalybu (changes requested):

README.md
mstrelex: @vitalybu better now ?

README.md
```diff
@@ -563,11 +563,16 @@ Here's an example of how this can be done with Gradle in build.gradle:
 
 ```gradle
 jar {
+    // Collect and zip all classes from both compile and test configurations
     from { configurations.testCompile.collect { it.isDirectory() ? it : zipTree(it) } }
```
vitalybu, line 567: This is used for packaging?
If so, maybe the order of the two lines should be inverted.
First, compile the test classes and then package everything into a zip.

mstrelex: I moved it, however it's not a compilation instruction but packaging that happens after the compilation is done.
Also updated `testCompile` to `testRuntimeClasspath`, otherwise the dependencies aren't packaged.

README.md
```diff
@@ -550,6 +550,26 @@ in your program's Run/Debug configuration as :
 
     If executing without a runner, it is possible to initialize
     the driver as part of the step definition class constructor.
+  
+# Package & Upload Tests to TestProject
+
+Tests can be executed locally using the SDK and remotely from the platform. \
```
vitalybu, line 556:
```suggestion
Tests can be executed locally using the SDK, or triggered remotely from the TestProject platform. \
```

vitalybu (approved):
Added one last comment, but approving to not block you from merging

PR #82, opened 2021-02-09
Avoid NullPointerException when reporting Cucumber step.

DavidG8168:
When there is no step error, and the step failed, getting the message from NULL will cause a NullPointerException.
Checking if the message is NULL before attempting to get the message, if it is, use "" as the message.


mstrelex (approved):

mstrelex:
@DavidG8168 Please prepare version-bump PR

PR #87, opened 2021-02-11
Avoid null pointer exception when creating Generic driver in Cucumber test.

DavidG8168:
If running Cucumber, when creating the reporter, it will attempt to disable the auto test and command reports,
it will then cause a null pointer exception when the reporting command executor has not been created yet when
using a Generic driver.


DavidG8168:
There are other ways to handle this, like checking if the driver is generic and not disabling auto tests and command reports, since it doesn't have them, or checking if the executor is null before disabling the tests and command reports.

vitalybu (approved):

PR #91, opened 2021-03-24
Allow to control report type - local / cloud or both

mstrelex:
When a **local** report is not supported - SDK will output:

`Target Agent version [2.0.0] doesn't support local reports. Upgrade the Agent to the latest version and try again.`

When local **cloud** is requested but Agent is offline, SDK will output:

`Agent responded with status 412: [Agent is offline and is not able to upload the execution report to the cloud as requested]`

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/rest/ReportSettings.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/ReportSettings.java
@@ -17,6 +17,8 @@
 
 package io.testproject.sdk.internal.rest;
 
+import io.testproject.sdk.drivers.ReportType;
+
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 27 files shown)


mstrelex:
@vitalybu Please review.

src/main/java/io/testproject/sdk/internal/rest/ReportSettings.java
```diff
@@ -60,6 +76,20 @@ public String getJobName() {
     public ReportSettings(final String projectName, final String jobName) {
         this.projectName = projectName;
         this.jobName = jobName;
+        this.reportType = ReportType.CLOUD_AND_LOCAL;
```
vitalybu, line 79: If `CLOUD_AND_LOCAL` is the default, what will happen when the Agent is offline?

mstrelex: It creates only a local report and shows a warning.

vitalybu (approved):

PR #96, opened 2021-04-04
Do not report methods without @Test annotation as JUnit tests.

DavidG8168:
Check method being reported when inferring JUnit test name for Test annotation, and return null to avoid
reporting it as a test if it does not have one.

Signed-off-by: David Goichman <david.goichman@testproject.io>

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/reporting/inferrers/JUnitInferrer.java
+++ b/src/main/java/io/testproject/sdk/internal/reporting/inferrers/JUnitInferrer.java
@@ -164,6 +164,15 @@ public class JUnitInferrer implements ReportSettingsInferrer {
                 continue;
             }
 
+            // If not JUnit @Test method, nothing to infer.
+            if (Arrays.stream(method.get().getDeclaredAnnotations())
+                    .noneMatch(a -> Arrays.asList(JUNIT4_TEST_ANNOTATION, JUNIT5_TEST_ANNOTATION)
+                                           .contains(a.annotationType().getName()))) {
+                LOG.trace("No JUnit test annotations detected, skipping inferring {} method name", method.get()
+                        .getName());
+                return null;
+            }
+
             // Search for @DisplayName annotation
             Optional<Annotation> displayNameAnnotation =
                     Arrays.stream(method.get().getDeclaredAnnotations())


src/main/java/io/testproject/sdk/internal/reporting/inferrers/JUnitInferrer.java
```diff
@@ -164,6 +164,17 @@ public String inferTestName() {
                 continue;
             }
 
+            // Search for JUnit @Test annotation
+            Optional<Annotation> testAnnotation =
+                    Arrays.stream(method.get().getDeclaredAnnotations())
+                            .filter(a -> a.annotationType().getName().equals(JUNIT5_TEST_ANNOTATION))
```
mstrelex, line 170: Is this applicable for JUnit5 only and not for JUnit4?

DavidG8168: @mstrelex Right, should consider JUnit4 as well, changed it to consider the JUnit4 test annotation as well.

src/main/java/io/testproject/sdk/internal/reporting/inferrers/JUnitInferrer.java
```diff
@@ -164,6 +165,18 @@ public String inferTestName() {
                 continue;
             }
 
+            // Search for JUnit @Test annotation
+            Predicate<Annotation> isJunit5 = a -> a.annotationType().getName().equals(JUNIT5_TEST_ANNOTATION);
+            Predicate<Annotation> isJunit4 = a -> a.annotationType().getName().equals(JUNIT4_TEST_ANNOTATION);
+            Optional<Annotation> testAnnotation = Arrays.stream(method.get().getDeclaredAnnotations())
+                    .filter(isJunit5.or(isJunit4))
```
mstrelex, line 172: Why not just `anyMatch()` instead of `filter()` and `findFirst()` ?

src/main/java/io/testproject/sdk/internal/reporting/inferrers/JUnitInferrer.java
```diff
@@ -164,6 +165,18 @@ public String inferTestName() {
                 continue;
             }
 
+            // Search for JUnit @Test annotation
+            Predicate<Annotation> isJunit5 = a -> a.annotationType().getName().equals(JUNIT5_TEST_ANNOTATION);
```
mstrelex, line 169: Use `Arrays.asList(JUNIT5_TEST_ANNOTATION, JUNIT5_TEST_ANNOTATION)` and the contains() for the next test.

DavidG8168:
@mstrelex It has been done.

mstrelex (approved):

PR #97, opened 2021-04-05
Add support for JUnit4 assertion reporting.

DavidG8168:
Must be implemented by extending the RunListener class and overriding it's failure method.
Then the listener must be registered in a runner class.

To enable the actual assertion reporting, the user will need to register the runner with the RunWith annotation in his test class, similar to the CucumberReporter which also works with JUnit4.

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/reporting/extensions/junit4/ExceptionsReporter.java
+++ b/src/main/java/io/testproject/sdk/internal/reporting/extensions/junit4/ExceptionsReporter.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2021 TestProject LTD. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


src/main/java/io/testproject/sdk/internal/reporting/extensions/junit4/ExceptionsReporter.java
```diff
@@ -0,0 +1,78 @@
...
+
+        // Report the error.
+        // Proceed only if error is instance of assertion error.
+        if (!(failure.getException() instanceof AssertionError)) {
+            return;
+        }
+
+        // Get the assertion error message.
+        String resultDescription = failure.getException().getMessage();
+        // Proceed only if message is not empty.
```
mstrelex, line 60: Do all assertions have messages? Even such as `assertNull`?

DavidG8168: @mstrelex `assertNull` does have a message:
`expected null, but was:<.....>`

From briefly going over all the Asserts, it looks like most if not all of them have their own failure messages and allow providing a custom message as well if necessary.

mstrelex (approved):

Issue #101, opened 2021-04-13
Running Cucumber features with TestNG creates an additional empty test in the report


s-glatshtein:
I created a cucumber test with the following feature:
```
Feature: TestProject with Cucumber Framework
  Perform a login scenario with Cucumber

  Scenario: Run a Simple BDD test with TestProject
    Given I navigate to the TestProject example page
    When I perform a login
    Then I should see a logout button
    And I should close the browser
```

This is my step definitions file:
```java
package mytests;

import io.cucumber.java.en.And;
import io.cucumber.java.en.Given;
import io.cucumber.java.en.Then;
import io.cucumber.java.en.When;
import io.testproject.sdk.drivers.web.ChromeDriver;
import org.junit.Assert;
import org.openqa.selenium.By;

import java.util.concurrent.TimeUnit;

/**
 * Step definitions for login_scenario.feature.
 */
public class StepDefinitions {

    /**
     * Driver used in this web test.
     */
    public static ChromeDriver driver;

    /**
     * Constructs the TestProject driver and navigates to the TestProject
     * example page.
     *
     */
    @Given("I navigate to the TestProject example page")
    public void navigateToPage() {
        final int timeout = 1500;
        driver.manage().timeouts().implicitlyWait(timeout, TimeUnit.MILLISECONDS);
        driver.navigate().to("https://example.testproject.io/web/");
    }

    /**
     * Performs a login in the TestProject example page.
     */
    @When("I perform a login")
    public void performLogin() {
        driver.findElement(By.cssSelector("#name")).sendKeys("John Smith");
        driver.findElement(By.cssSelector("#password")).sendKeys("12345");
        driver.findElement(By.cssSelector("#login")).click();
    }

    /**
     * Validates if the login was successful by checking if the login button appears.
     */
    @Then("I should see a logout button")
    public void validateLogoutButton() {
        Assert.assertTrue(driver.findElement(By.cssSelector("#logout")).isDisplayed());
    }

    /**
     * Quits the driver once and closes the session.
     */
    @And("I should close the browser")
    public void closeBrowser() {
       // Do nothing for now
    }
}
```

I use TestNG to run the tests. I create the driver before all scenarios and quit it after:
```java
import io.cucumber.testng.AbstractTestNGCucumberTests;
import io.cucumber.testng.CucumberOptions;
import io.testproject.sdk.drivers.web.ChromeDriver;
import io.testproject.sdk.internal.exceptions.AgentConnectException;
[...]

PR #124, opened 2021-05-10
Fix: Disable reports for Cucumber tests

Rantzur1992:
This is a fix for https://github.com/testproject-io/java-opensdk/issues/101

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
+++ b/src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
@@ -40,6 +40,11 @@ public final class Reporter {
     private static final Logger LOG = LoggerFactory.getLogger(Reporter.class);
 
     /**
+     * Regex which checks if the class is a Cucumber runner class.
+     */
+    private static final String CUCUMBER_OPTIONS_REGEX = "io\\.cucumber\\.\\w+\\.CucumberOptions";
+
+    /**
      * {@link AgentClient} instance to submit reports to the Agent.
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -90,6 +90,9 @@ dependencies {
 
     // Apache commons
     implementation group: 'org.apache.commons', name: 'commons-csv', version: '1.8'
+
+    implementation group: 'io.cucumber', name: 'cucumber-testng', version: '6.10.3'
```
s-glatshtein, line 94: just drop it!
make your dreams come true!

src/main/java/io/testproject/sdk/internal/reporting/Reporter.java
```diff
@@ -58,12 +63,53 @@
...
+    }
+
+    /**
+     * Checks if the current class is annotated with a cucumber annotation.
+     * @param callerClass The current class.
+     * @return True if the class is annotated with CucumberOptions, false otherwise.
+     */
+    private boolean isCucumberAnnotationPresent(final Class<?> callerClass) {
+        return callerClass != null && Arrays.stream(callerClass.getDeclaredAnnotations()).anyMatch(a ->
+                a.annotationType().getName().contains(CUCUMBER_OPTIONS));
```
s-glatshtein, line 93: should check for the actual name (can be a regex like `io.<>.CucumberOptions` or w/e the canonical names both match

mstrelex (approved):

PR #137, opened 2021-06-10
Increase maximum reporting time and log reporting progress,

DavidG8168:
- Increased the maximum time for the queue to 10 minutes.
- Gave each queue item a unique index, logged it after it was reported along with the amount of reports remaining in the queue.


s-glatshtein:
@DavidG8168 while we are waiting for queue to clear, we should write to log how many items are left every few seconds or so

DavidG8168:
@s-glatshtein implemented a scheduled future to do so.

mstrelex:
@DavidG8168 

In the second commit 3abb204aa21bc0506e9310f2ce664c40556b5a95, there is a typo: `reamining`

DavidG8168:
@mstrelex fixed the typo.

PR #140, opened 2021-06-17
Prioritize Development Token Provided in TP_DEV_TOKEN variable over token in constructor.

DavidG8168:


README.md
```diff
@@ -142,6 +142,8 @@ Token can be also provided explicitly using this constructor:
 public ChromeDriver(final String token, final ChromeOptions options)
 ```
 
+**NOTE:** When a token is provided in both the constructor and an environment variable, the token in the environment variable will be used.
```
mstrelex, line 145: Notes are usually preceded with `>`, e.g.

> This is a note.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -240,10 +240,11 @@ private AgentClient(final URL remoteAddress, final String token, final Capabilit
         this.remoteAddress = inferRemoteAddress(remoteAddress);
 
         // Determine Development Token
-        if (!StringUtils.isEmpty(token)) {
-            this.token = token;
-        } else if (!StringUtils.isEmpty(System.getenv(TP_DEV_TOKEN))) {
+        // Prioritize token provided in environment variable
+        if (!StringUtils.isEmpty(System.getenv(TP_DEV_TOKEN))) {
             this.token = System.getenv(TP_DEV_TOKEN);
```
s-glatshtein, line 245: Add a log if user supplied dev token parameter, to tell them we are using token from environment instead of the parameter

s-glatshtein:
@DavidG8168 FYI there is a `docs` prefix that should be  used when updating  documentation

mstrelex (approved):

Issue #145, opened 2021-07-01
Reports performance with slow internet connection - allow reporting steps at batches from OpenSDK


ori-d:
Due to performance issues, with slow internet connection. 
Current logic is reporting steps one by one. 
Additional logic is required to fix that performance issues with slow internet connection.
Instead of reporting steps one by one - report steps at batches of up to 10 steps in a time.

mstrelex:
@ori-d I'm not sure it has anything to do with internet connectivity, but with the latency between the machine where the SDK is running the test and the Agent running on a remote machine, accepting these steps reports...

PR #146, opened 2021-07-05
Add new ReportsQueue to support reporting at batches.

oriddd:
Adding logic to support reporting at batches due to issue #145 . 
ReportsQueueBatch was added to support up to 10 reports in a batch. 
Logic added to AgentClient - to support the new ReportsQueue from agent version: 3.0.0

### DIFF
--- a/src/main/java/io/testproject/sdk/internal/exceptions/FailedReportException.java
+++ b/src/main/java/io/testproject/sdk/internal/exceptions/FailedReportException.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2020 TestProject LTD. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.testproject.sdk.internal.exceptions;
+
+/**
+ * This exception is thrown when SDK fails to send a report to the agent.
+ */
+public class FailedReportException extends Exception {
+    /**
+     * Default serial version ID.
+     */
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Creates a new instance of the class with a default message.
+     */
+    public FailedReportException() {
+        this("Failed sending a report to the Agent");
+    }
+
+    /**
+     * Creates a new instance of the class with a provided message.
+     *
+     * @param message message to be set
+     */
+    public FailedReportException(final String message) {
+        super(message);
+    }
+
+    /**
+     * Creates a new instance of the class with a provided message and cause.
+     *
+     * @param message message to be set
+     * @param cause   another exception that caused this one
+     */
+    public FailedReportException(final String message, final Throwable cause) {
+        super(message, cause);
+    }
+}
--- a/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
@@ -1236,7 +1236,7 @@ public final class AgentClient implements Closeable {
     /**
      * Internal class used to store Agent API routes.
      */
-    private static class Routes {
+    static class Routes {
         /**
          * Agent "status" endpoint address.
          */
@@ -1261,6 +1261,11 @@ public final class AgentClient implements Closeable {
          * Test reporting endpoint address.
          */
         static final String REPORT_TEST = "/api/development/report/test";
+
+        /**
+         * Batch reporting endpoint address.
+         */
+        static final String REPORT_BATCH = "/api/development/report/batch";
 
         /**
          * Action proxy execution endpoint address.
--- a/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
@@ -211,6 +211,12 @@ public final class AgentClient implements Closeable {
     private SessionResponse agentResponse;
 
     /**
+     * Minimum Agent version that supports batch reporting.
+     */
+    private static final String MIN_BATCH_REPORT_SUPPORTED_VERSION = "3.1.0";
+
+
+    /**
      * Creates a new instance of the class.
      * Initiates a development session with the Agent.
      *
@@ -290,7 +296,13 @@ public final class AgentClient implements Closeable {
 
         // Start reports queue
         if (!disableReports) {
-            this.reportsQueue = new ReportsQueue(this.httpClient, this.getSession().getSessionId());
+            if (new ComparableVersion(version).compareTo(
+                    new ComparableVersion(MIN_BATCH_REPORT_SUPPORTED_VERSION)) >= 0) {
+                        this.reportsQueue = new ReportsQueueBatch(
+                                this.httpClient, this.getSession().getSessionId(), this.remoteAddress);
+            } else {
+                this.reportsQueue = new ReportsQueue(this.httpClient, this.getSession().getSessionId());
+            }
             this.reportsQueueFuture = reportsExecutorService.submit(this.reportsQueue);
         }
 
--- a/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
@@ -19,23 +19,24 @@ package io.testproject.sdk.internal.rest;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import io.testproject.sdk.internal.exceptions.FailedReportException;
 import io.testproject.sdk.internal.rest.messages.Report;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
+import org.apache.http.client.methods.HttpPost;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.util.EntityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.ws.rs.core.Response;
 import java.io.IOException;
-import java.net.HttpURLConnection;
 import java.util.concurrent.*;
 
 /**
  * A runnable class to manage reports queue.
  */
 public class ReportsQueue implements Runnable {
-
     /**
      * Logger instance.
      */
@@ -44,10 +45,22 @@ public class ReportsQueue implements Runnable {
     /**
      * An instance of the Google JSON serializer to serialize and deserialize objects.
      */
-    private static final Gson GSON = new GsonBuilder().create();
+    protected static final Gson GSON = new GsonBuilder().create();
+
+    /**
+     * Progress report delay in seconds.
+     * Print the number of the remaining reports to send every 3 seconds.
+     */
+    private static final int PROGRESS_REPORT_DELAY = 3;
+
+    /**
+     * Maximum attempts to try sending a report to the Agent.
+     */
+    protected static final int MAX_REPORT_FAILURE_ATTEMPTS = 4;
 
     /**
      * Queue to synchronize reports sent to Agent.
+     * Queue depth defined as 10K - assuming that even with very high latency, the queue won't get full.
      */
     private final LinkedBlockingQueue<QueueItem> queue = new LinkedBlockingQueue<>(1024 * 10);
 
@@ -72,9 +85,9 @@ public class ReportsQueue implements Runnable {
     private Future<?> progressFuture;
 
     /**
-     * Progress report delay in seconds.
+     * A flag that is raised when all attempts submitting a report fail.
      */
-    private static final int PROGRESS_REPORT_DELAY = 3;
+    private boolean stopReports = false;
 
     /**
      * Initializes a new instance of the class.
@@ -94,7 +107,30 @@ public class ReportsQueue implements Runnable {
      * @param report  Report that this request contains.
      */
     void submit(final HttpEntityEnclosingRequestBase request, final Report report) {
-        this.queue.add(new QueueItem(request, report));
+        if (!this.stopReports) {
+            this.queue.add(new QueueItem(request, report));
+        }
+    }
+
+    /**
+     * For lower versions than 3.1.0 -> Send standalone reports.
+     * @throws InterruptedException in case reports queue was interrupted
+     * @throws FailedReportException in case of 4 failures to send reports to the agent
+     */
+    void handleReport() throws InterruptedException, FailedReportException {
+        QueueItem item = this.queue.take();
+
+        if (item.getRequest() == null && item.getReport() == null) {
+            if (this.running) {
+                // There nulls are not OK, something went wrong preparing the report/request.
+                LOG.error("Empty report and request were submitted to the queue!");
+            }
+
+            // These nulls are OK - they were added by stop() method on purpose.
+            return;
+        }
+
+        sendReport((HttpPost) item.getRequest());
     }
 
     /**
@@ -103,13 +139,15 @@ public class ReportsQueue implements Runnable {
     @Override
     public void run() {
         this.running = true;
-
-        while (this.running || queue.size() > 0) {
+        while (this.running || !this.queue.isEmpty()) {
             try {
-                sendReport(queue.take());
+                handleReport();
             } catch (InterruptedException e) {
                 LOG.error("Reports queue was interrupted");
                 break;
+            } catch (FailedReportException e) {
+                this.stopReports = true;
+                LOG.warn("Reports are disabled due to multiple failed attempts of sending reports to the agent.");
             }
         }
 
@@ -122,38 +160,47 @@ public class ReportsQueue implements Runnable {
 
     /**
      * Submits a report to the Agent via HTTP RESTFul API endpoint.
-     *
-     * @param item Item retrieved from the queue (report & HTTP request).
+     * @param httpPost For lower versions than 3.1.0 -> HTTP request retrieved from the queue.
+     *                 For versions 3.1.0 and greater -> Build reports batch HTTP request.
+     * @throws FailedReportException in case of 4 failures to send reports to the agent
      */
-    private void sendReport(final QueueItem item) {
-        if (item.getRequest() == null && item.getReport() == null) {
-            if (this.running) {
-                // There nulls are not OK, something went wrong preparing the report/request.
-                LOG.error("Empty report and request were submitted to the queue!");
-            }
-
-            // These nulls are OK - they were added by stop() method on purpose.
-            return;
-        }
-
+    void sendReport(final HttpPost httpPost) throws FailedReportException {
+        int reportAttemptsCount;
         CloseableHttpResponse response = null;
-        try {
-            response = this.httpClient.execute(item.getRequest());
-        } catch (IOException e) {
-            LOG.error("Failed to submit report: [{}]", item.getReport(), e);
-            return;
-        } finally {
-            // Consume response to release the resources
+        // Send the report to the agent.
+        // In case of failure - make 3 more attempts.
+        for (reportAttemptsCount = MAX_REPORT_FAILURE_ATTEMPTS; reportAttemptsCount > 0; reportAttemptsCount--) {
+            try {
+                response = this.getHttpClient().execute(httpPost);
+            } catch (IOException e) {
+                LOG.error("Failed to submit report.", e);
+            } finally {
+                // Consume response to release the resources
+                if (response != null) {
+                    EntityUtils.consumeQuietly(response.getEntity());
+                }
+            }
+
             if (response != null) {
-                EntityUtils.consumeQuietly(response.getEntity());
+                // If the reports were sent successfully, there is no need to continue to the rest of the code
+                // since it's handling unsuccessful response.
+                if (Response.Status.Family.familyOf(response.getStatusLine().getStatusCode())
+                    == Response.Status.Family.SUCCESSFUL) {
+                    return;
+                }
+
+                // Handle unsuccessful response
+                LOG.warn("Agent responded with an unexpected status {} to report.",
+                        response.getStatusLine().getStatusCode());
+                LOG.info("Failed to send a report to the Agent, {} attempts remaining...", reportAttemptsCount - 1);
             }
         }
 
-        // Handle unsuccessful response
-        if (response != null && response.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
-            LOG.error("Agent responded with an unexpected status {} to report: [{}]",
-                    response.getStatusLine().getStatusCode(), item.getReport());
-        }
+        // In case all attempts to send the report are failed.
+        // If the report has been sent successfully - this code will not execute.
+        LOG.error("All {} attempts to send report have failed.", MAX_REPORT_FAILURE_ATTEMPTS);
+        throw new FailedReportException("All " + MAX_REPORT_FAILURE_ATTEMPTS
+                + " attempts to send report have failed.");
     }
 
     /**
--- a/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
@@ -101,6 +101,22 @@ public class ReportsQueue implements Runnable {
     }
 
     /**
+     * Getter method for {@link #queue}.
+     * @return the reports queue.
+     */
+    protected LinkedBlockingQueue<QueueItem> getQueue() {
+        return queue;
+    }
+
+    /**
+     * Getter method for {@link #httpClient}.
+     * @return the httpClient.
+     */
+    protected CloseableHttpClient getHttpClient() {
+        return httpClient;
+    }
+
+    /**
      * Adds a report to the queue.
      *
      * @param request Request to be sent over HTTP.
--- a/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
@@ -49,7 +49,7 @@ public class ReportsQueue implements Runnable {
     /**
      * Queue to synchronize reports sent to Agent.
      */
-    private final ArrayBlockingQueue<QueueItem> queue = new ArrayBlockingQueue<>(1024 * 1024);
+    private final LinkedBlockingQueue<QueueItem> queue = new LinkedBlockingQueue<>(1024 * 10);
 
     /**
      * HTTP client to submit reports to the Agent.
--- a/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
@@ -188,7 +188,7 @@ public class ReportsQueue implements Runnable {
     /**
      * Internal class to keep the HTTP request and contained report together in the queue.
      */
-    private static class QueueItem {
+    static class QueueItem {
 
         /**
          * HTTP request.
--- a/src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright (c) 2020 TestProject LTD. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.testproject.sdk.internal.rest;
+
+import io.testproject.sdk.internal.exceptions.FailedReportException;
+import io.testproject.sdk.internal.rest.messages.Report;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.LinkedList;
+import java.util.List;
+
+public class ReportsQueueBatch extends ReportsQueue {
+    /**
+     * Logger instance.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(ReportsQueueBatch.class);
+
+    /**
+     * remote url to send the reports to.
+     */
+    private final URL remoteAddress;
+
+    /**
+     * The default batch report size is a maximum of 10 reports.
+     */
+    private static final int MAX_REPORTS_BATCH_SIZE = 10;
+
+    /**
+     * Constant for environment variable name that may store the max batch size.
+     */
+    private static final String TP_MAX_BATCH_SIZE_VARIABLE_NAME = "TP_MAX_REPORTS_BATCH_SIZE";
+
+    /**
+     * Class member to store actual max batch size.
+     */
+    private final int maxBatchSize;
+
+    /**
+     * Initializes a new instance of the class.
+     *
+     * @param httpClient HTTP client ot use for communicating with the Agent.
+     * @param sessionId  Driver session ID.
+     * @param remoteAddress Reports remote address.
+     */
+    ReportsQueueBatch(final CloseableHttpClient httpClient, final String sessionId, final URL remoteAddress) {
+        super(httpClient, sessionId);
+        this.remoteAddress = remoteAddress;
+
+        // Try to get maximum report batch size from env variable.
+        int maxBatchSizeValueToSet;
+        try {
+            String tpMaxBatchSizeEnvVarValue = System.getenv(TP_MAX_BATCH_SIZE_VARIABLE_NAME);
+            maxBatchSizeValueToSet = (tpMaxBatchSizeEnvVarValue != null)
+                    ? Integer.parseInt(tpMaxBatchSizeEnvVarValue) : MAX_REPORTS_BATCH_SIZE;
+        } catch (SecurityException e) {
+            LOG.warn("Failed to retrieve the value of environment variable {}. "
+                            + "Maximum reports batch size is set to the default value: {}.",
+                    TP_MAX_BATCH_SIZE_VARIABLE_NAME, MAX_REPORTS_BATCH_SIZE, e);
+            maxBatchSizeValueToSet = MAX_REPORTS_BATCH_SIZE;
+        } catch (Exception e) {
+            LOG.warn("Failed to convert the value of environment variable {}. "
+                            + "Maximum reports batch size is set to the default value: {}.",
+                    TP_MAX_BATCH_SIZE_VARIABLE_NAME, MAX_REPORTS_BATCH_SIZE, e);
+            maxBatchSizeValueToSet = MAX_REPORTS_BATCH_SIZE;
+        }
+        this.maxBatchSize = maxBatchSizeValueToSet;
+    }
+
+    /**
+     * From version 3.1.0 -> send reports in batches.
+     * Collect reports from reports queue and build reports batch.
+     * Send the reports batch to the agent.
+     * @throws InterruptedException in case reports queue was interrupted
+     * @throws FailedReportException in case of 4 failures to send reports to the agent
+     */
+    @Override
+    void handleReport() throws InterruptedException, FailedReportException {
+        // Linked list to store the reports batch before sending them.
+        // Those reports will be extract from the queue.
+        List<Report> batchReports = new LinkedList<>();
+
+        // Extract and remove up to {this.maxBatchSize} items or till queue is empty from queue - without blocking it.
+        while (!getQueue().isEmpty() && batchReports.size() < this.maxBatchSize) {
+            // Get the first item in the queue without blocking it.
+            QueueItem item = getQueue().poll();
+
+            if (null != item && item.getReport() != null) {
+                batchReports.add(item.getReport());
+            }
+        }
+
+        if (batchReports.isEmpty()) {
+            return;
+        }
+
+        // Create json from reports list
+        String json = null;
+        try {
+            json = this.GSON.toJson(batchReports);
+        } catch (Exception e) {
+            LOG.error("Failed to create json from list: [{}]", batchReports, e);
+            return;
+        }
+
+        // Parse batch report to json string
+        StringEntity entity = new StringEntity(json, StandardCharsets.UTF_8);
+
+        // Create httpPost and execute
+        HttpPost httpPost = new HttpPost(this.remoteAddress + AgentClient.Routes.REPORT_BATCH);
+        httpPost.setEntity(entity);
+
+        this.sendReport(httpPost);
+    }
+}
--- a/src/main/java/io/testproject/sdk/internal/rest/messages/DriverCommandReport.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/messages/DriverCommandReport.java
@@ -49,6 +49,11 @@ public final class DriverCommandReport extends Report {
     private String screenshot;
 
     /**
+     * Define type as Command for batch report support.
+     */
+    private final ReportItemType type = ReportItemType.Command;
+
+    /**
      * Getter for {@link #screenshot} field.
      *
      * @return value of {@link #screenshot} field
@@ -125,4 +130,5 @@ public final class DriverCommandReport extends Report {
     public String toString() {
         return this.commandName;
     }
+
 }
--- a/src/main/java/io/testproject/sdk/internal/rest/messages/ReportItemType.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/messages/ReportItemType.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright (c) 2020 TestProject LTD. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.testproject.sdk.internal.rest.messages;
+
+public enum ReportItemType {
+    /**
+     * Those are the reports types which are supported to be sent at batches.
+     */
+    Command, Test, Step
+}
--- a/build.gradle
+++ b/build.gradle
@@ -90,6 +90,9 @@ dependencies {
 
     // Apache commons
     implementation group: 'org.apache.commons', name: 'commons-csv', version: '1.8'
+
+    // Java API for RESTful Web Services
+    implementation 'javax.ws.rs:javax.ws.rs-api:2.1.1'
 }
 
 checkstyle.configFile file("config/checkstyle/checkstyle.xml")
--- a/src/main/java/io/testproject/sdk/internal/rest/messages/StepReport.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/messages/StepReport.java
@@ -17,6 +17,7 @@
 
 package io.testproject.sdk.internal.rest.messages;
 
+
 import java.util.UUID;
 
 /**
@@ -48,6 +49,12 @@ public final class StepReport extends Report {
      * Flag to indicate pass/fail state.
      */
     private final boolean passed;
+
+    /**
+     * Define type as Step for batch report support.
+     */
+    private final ReportItemType type = ReportItemType.Step;
+
 
     /**
      * Getter for {@link #guid} field.
@@ -120,4 +127,5 @@ public final class StepReport extends Report {
     public String toString() {
         return this.description;
     }
+
 }
--- a/src/main/java/io/testproject/sdk/internal/rest/messages/TestReport.java
+++ b/src/main/java/io/testproject/sdk/internal/rest/messages/TestReport.java
@@ -38,6 +38,11 @@ public final class TestReport extends Report {
     private String message;
 
     /**
+     * Define type as Test for batch report support.
+     */
+    private final ReportItemType type = ReportItemType.Test;
+
+    /**
      * Initializes a new instance of a Test Report.
      *
      * @param name Test name.
@@ -121,4 +126,5 @@ public final class TestReport extends Report {
     public String toString() {
         return this.name;
     }
+
 }


mstrelex:
@oriddd 

Can you please change the PR title to reflect the actual changes?

src/main/java/io/testproject/sdk/internal/rest/messages/Report.java
```diff
@@ -17,5 +17,29 @@
 
 package io.testproject.sdk.internal.rest.messages;
 
-public class Report {
+public abstract class Report {
+
+    /**
+     * Type of the report.
+     * Should be Command / Test / Step.
```
mstrelex, line 24: Why not an enum if the list is closed?
And thus the classes `DriverCommandReport`, `StepReport`, and `TestReport` could define it as final constants.

src/main/java/io/testproject/sdk/internal/rest/messages/Report.java
```diff
@@ -17,5 +17,29 @@
...
+     */
+    private String type;
+
+    /**
+     * Used for json serialization.
```
mstrelex, line 29: GSON can serialize private fields, why a getter is required?

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -210,6 +210,12 @@
      */
     private SessionResponse agentResponse;
 
+    /**
+     * Minimum Agent version that supports local reports.
+     */
+    private static final String MIN_BATCH_REPORT_SUPPORTED_VERSION = "3.0.0";
```
mstrelex, line 216: Should be `3.1.0` as this is the version where it will be introduced first.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.concurrent.*;
+
+public abstract class ReportsQueueBase implements Runnable {
+    /**
+     * Logger instance.
+     */
+    static final Logger LOG = LoggerFactory.getLogger(ReportsQueue.class);
```
mstrelex, line 38: ReportsQueue --> ReportsQueueBase

mstrelex: Should have been split into a separate `refactor` commit to make it easier to understand whether there were any changes or not...

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,119 @@
...
+                    batchReports.add(item.getReport());
+                }
+            } while (null != item && batchReports.size() < MAX_REPORTS_BATCH_SIZE);
+
+            // Create json from reports list
+            String json = null;
+            try {
+                json = GSON.toJson(batchReports);
+            } catch (Exception e) {
+                LOG.error("Failed to create json from list: [{}]", batchReports, e);
```
mstrelex, line 87: If it failed, shouldn't the method return?

mstrelex (changes requested):

src/main/java/io/testproject/sdk/internal/rest/messages/DriverCommandReport.java
```diff
@@ -48,6 +48,12 @@
      */
     private String screenshot;
 
+    /**
+     * Define type as Command for batch report support.
+     */
+    private final String type = ReportItemType.Command.toString();
```
mstrelex, line 54: Still, why a string and not an enum?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
+/*
+ * Copyright (c) 2020 TestProject LTD. and/or its affiliates
```
vitalybu, line 2: 2021

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.util.EntityUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.concurrent.*;
```
vitalybu, line 32: Do you really need all the classes from this package?
Usually it's better to import only what you actually use.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * An instance of the Google JSON serializer to serialize and deserialize objects.
+     */
+    static final Gson GSON = new GsonBuilder().create();
+
+    /**
+     * Queue to synchronize reports sent to Agent.
+     */
+    private final ArrayBlockingQueue<ReportsQueue.QueueItem> queue = new ArrayBlockingQueue<>(1024 * 1024);
```
vitalybu, line 48: I think we already had a discussion about the size of this queue.
We don't really need an array with a million items, especially since it's pre-allocated.
@mstrelex, do you remember what was the decision?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.concurrent.*;
+
+public abstract class ReportsQueueBase implements Runnable {
+    /**
+     * Logger instance.
+     */
+    static final Logger LOG = LoggerFactory.getLogger(ReportsQueueBase.class);
```
vitalybu, line 38: `private static final`?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * An instance of the Google JSON serializer to serialize and deserialize objects.
+     */
+    static final Gson GSON = new GsonBuilder().create();
```
vitalybu, line 43: `private static final`?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Future to report remaining reports in queue.
+     */
+    private Future<?> progressFuture;
+
+    /**
+     * Progress report delay in seconds.
+     */
+    private static final int PROGRESS_REPORT_DELAY = 3;
```
vitalybu, line 73: Final members should be declared before any other members.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+     * Adds a report to the queue.
+     *
+     * @param request Request to be sent over HTTP.
+     * @param report  Report that this request contains.
+     */
+     void submit(final HttpEntityEnclosingRequestBase request, final Report report) {
+        this.queue.add(new QueueItem(request, report));
+    }
+
+
```
vitalybu, line 112: Remove extra new line

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * Handle the report.
+     * Till version x -> send standalone report.
```
vitalybu, line 115: `version x`?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+     */
+    abstract void handleReport() throws InterruptedException;
+
+    /**
+     * Runnable flow that looks into the queue and waits for new items.
+     */
+    @Override
+    public void run() {
+        this.running = true;
+        while (this.running || queue.size() > 0) {
```
vitalybu, line 127: Why not `!this.queue.isEmpty()`?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    /**
+     * Submits a report to the Agent via HTTP RESTFul API endpoint.
+     *
+     * @param item Item retrieved from the queue (report & HTTP request).
+     */
+    void sendReport(final QueueItem item) {
+        if (item.getRequest() == null && item.getReport() == null) {
```
vitalybu, line 149: Only if both are `null` it's considered as error?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+            LOG.error("Failed to submit report: [{}]", item.getReport(), e);
+            return;
+        } finally {
+            // Consume response to release the resources
+            if (response != null) {
+                EntityUtils.consumeQuietly(response.getEntity());
+            }
+        }
+
+        // Handle unsuccessful response
```
vitalybu, line 172: Maybe we should stop reporting after a certain number of errors?
For example, if the Agent crashed for some reason, the SDK will keep trying to report everything...

vitalybu: Another case is if the Agent returns 401 or 403, this probably means that the development token is not valid, so there's no reason to continue reporting.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
```diff
@@ -0,0 +1,257 @@
...
+     * Stops the runnable and the queue processing.
+     */
+    public void stop() {
+        LOG.trace("Raising flag to stop reports queue for session [{}]", sessionId);
+        this.running = false;
+
+        // Feed the queue with one more (null) object.
+        // This is required to to let it proceed with the loop to evaluate the condition (running?) again.
+        // Note: Sending null as QueueItem is not possible since ArrayBlockingQueue prohibit null elements.
+        this.queue.add(new QueueItem(null, null));
```
vitalybu, line 190: I don't understand why this is required.
Also it doesn't make sense sending nulls if it's "not possible"

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -17,216 +17,31 @@
...
-     * @param report  Report that this request contains.
-     */
-    void submit(final HttpEntityEnclosingRequestBase request, final Report report) {
-        this.queue.add(new QueueItem(request, report));
-    }
-
-    /**
-     * Runnable flow that looks into the queue and waits for new items.
+     * Handle the report.
+     * Till version x -> send standalone report.
```
vitalybu, line 39: `version x`?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -17,216 +17,31 @@
...
-     * Future to report remaining reports in queue.
-     */
-    private Future<?> progressFuture;
-
-    /**
-     * Progress report delay in seconds.
-     */
-    private static final int PROGRESS_REPORT_DELAY = 3;
-
+public class ReportsQueue extends ReportsQueueBase {
```
vitalybu, line 26: This class doesn't do much, why do we need it?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright (c) 2020 TestProject LTD. and/or its affiliates
```
vitalybu, line 2: 2021

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * remote url to send the reports to.
+     */
+    private final URL remoteAddress;
+
+    /**
+     * Array list to store represent the reports batch before sending them.
+     * Those reports will be extract from the queue.
+     */
+    private List<Report> batchReports = new ArrayList<>();
```
vitalybu, line 45: Why `ArrayList` and not `LinkedList`?
Do we really array functionality?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Batch report size is maximum 10 reports.
+     */
+    private static final int MAX_REPORTS_BATCH_SIZE = 10;
```
vitalybu, line 50: Maybe we should allow overriding the default using an environment variable (and/or some SDK option)

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    ReportsQueueBatch(final CloseableHttpClient httpClient, final String sessionId, final URL remoteAddress) {
+        super(httpClient, sessionId);
+        this.remoteAddress = remoteAddress;
+    }
+
+
+    /**
+     * Handle the report.
+     * Till version x -> send standalone report.
```
vitalybu, line 67: `version x`?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,120 @@
...
+     * From version x -> send reports in batches.
+     */
+    @Override
+    void handleReport() throws InterruptedException {
+        // extract and remove up to 10 items or till queue is empty from queue - without blocking it.
```
vitalybu, line 72: Comments should start from a capital letter

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,120 @@
...
+        if (!getQueue().isEmpty()) {
+            QueueItem item = null;
+            do {
+                item = getQueue().poll();
+                if (null != item && batchReports.size() < MAX_REPORTS_BATCH_SIZE) {
+                    batchReports.add(item.getReport());
+                }
+            } while (null != item && batchReports.size() < MAX_REPORTS_BATCH_SIZE);
```
vitalybu, line 80: Why?
It should report whatever is in the queue, up to the maximum allowed items.
We shouldn't wait for the items to accumulate in the qeueu.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,120 @@
...
+                    EntityUtils.consumeQuietly(response.getEntity());
+                }
+                batchReports.clear();
+            }
+
+            // Handle unsuccessful response
+            if (response != null && response.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
+                LOG.error("Agent responded with an unexpected status {} to report: [{}]",
+                        response.getStatusLine().getStatusCode(), batchReports);
+            }
```
vitalybu, line 115: Previous comment for handling Agent response codes is relevant for this part as well

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -1262,6 +1262,11 @@ public void updateJobName(final String updatedJobName) {
          */
         static final String REPORT_TEST = "/api/development/report/test";
 
+        /**
+         * Test reporting endpoint address.
```
vitalybu, line 1278: Not sure the comment is correct here

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -210,6 +210,12 @@
      */
     private SessionResponse agentResponse;
 
+    /**
+     * Minimum Agent version that supports local reports.
```
vitalybu, line 214: ... supports **batch** reports?

vitalybu (changes requested):

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
oriddd: This behavior was exist. 
Once both are nulls and the queue is running -> this is an error.
The expected for both to be null is only once the queue is not running, this was defined to release the block for the queue.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
oriddd: This is the empty item that should release the queue.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
oriddd: This class is confirming that previous logic will still be exist.
In the original code - handleReport() method was not exist and the run() method was executed sendReport() method.
I've change the logic so that:
For version 3.1.0 and above -> handleReport() method will build ArrayList of reports and will send them as a batch.
For lower versions -> handleReport() method will perform the original logic.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
s-glatshtein: as the comment suggests, this will release the thread blocked by `queue.take()`. This is expected only when the queue is no longer running.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
vitalybu: And I left a comment on the "release" logic as well... I don't understand why it was designed this way.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
vitalybu: Can't we interrupt the thread instead of injecting an invalid value?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
vitalybu: But if the `ReportsQueue` doesn't actually add new logic, why not use the `ReportsQueueBase` for the old logic and override it in a new class for the new logic?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
mstrelex: I believe that when you work with IntelliJ, and you optimize the imports, it decides to put a wildcard after some specific threshold. This does not "include" the class, but just a way to tell Java where to find the definition of that class. There is no reason why not to put an asterisk, in order to reduce the number of imports.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
mstrelex: Same as my previous comment regarding the license header in general.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
mstrelex: This might be a better approach, but this is not a new code. It existed in this class before it was renamed, and unless you see some exotic negative consequences, I don't see a reason to modify it.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
mstrelex: https://stackoverflow.com/a/322742/712160

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
mstrelex: Maybe.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
mstrelex: Agree.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
mstrelex: Right, we should report anything there is or split it into chunks if it breaches the threshold.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: It's definitely not.

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
mstrelex: Right.

mstrelex (changes requested):

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
mstrelex: @vitalybu 

You said:

```
I'm not saying we need to have a queue with depth of 5 items, but 1k seems very reasonable to me. And no, returning an exception is not a good option, but maybe increasing the queue dynamically (up to some sane value) or even blocking the user in such a case, is a totally valid option.
If a user fills up the queue, maybe it's even better to block the calls to prevent flooding our backend.```

I answered:

```
Since there is a latency between the machine where the SDK is running and the Agent, this queue fills up, this is not the queue for reporting anything to the cloud but from the SDK to the Agent.
﻿﻿Indeed makes sense to refactor into LinkedBlockingQueue
﻿﻿﻿```
﻿﻿﻿
Previously we had a value of 1024 and it wasn't enough.
What would you like to specify here now?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
mstrelex: I don't see this comment, what did you write there?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
vitalybu: If we'll switch to a dynamically allocated queue, let's go with a maximal value of 10k (assuming 1k wasn't enough).

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
mstrelex: This can be addressed in a separate PR, as this is not the purpose here.

src/main/java/io/testproject/sdk/internal/rest/messages/ReportItemType.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2020 TestProject LTD. and/or its affiliates
```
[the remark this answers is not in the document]
vitalybu: Should be updated for new (or renamed) files as part of the PR modifying them...

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
vitalybu: You believe this is the case or is it really the case?
In any case, I'm not sure we want to rely on an IDE for this.
Importing an entire package can cause collisions with similar classes from different packages.
An IDE will probably detect and fix it automatically, but it's better to just import what is actually used.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
vitalybu: If this logic wasn't added as part of this PR you're right.
But at least an issue should be created for this to be handled later.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
vitalybu: The negative consequences are extra logic that eventually can cause other issues and code that is harder to read/review... 
Personally, I consider something that needs to be explained by a three-line-comment and can be avoided, bad design.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
s-glatshtein: Isn't this the case? either there's no item in queue (item is null) or the batch reaches 10 items (reportsSize == 10)

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
vitalybu: We use it as a simple queue (push to the end and pull from the start), so a linked list is much more appropriate here.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
vitalybu: Great, so let's do it

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
vitalybu: Maybe I'm unfamiliar with the queue syntax, but I understand this as if we're filling the `batchReports` with 10 items and only then report it to the Agent.
If the queue has only 5 items, I don't see a reason to wait for additional 5 to queue before reporting whatever is available to the Agent.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
vitalybu: That we need to properly handle error response codes from the Agent.
There's no reason to keep reporting if the Agent returns 401, 403 or 500 for that matter... Maybe other codes as well.

src/main/java/io/testproject/sdk/internal/rest/messages/ReportItemType.java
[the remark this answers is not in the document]
mstrelex: It can't work that way, checkstyle has only one template to compare with.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
mstrelex: I base my assumption on personal experience with IntelliJ.
I still don't understand what is the problem is everything compiles?
We can use a notepad that does not optimize anything on its own.

vitalybu: That's exactly the problem, sometimes it compiles but doesn't use the class we intended to use.
I remember more than one example of issues that were caused by using a class with the same name but from a different package.

src/main/java/io/testproject/sdk/internal/rest/messages/ReportItemType.java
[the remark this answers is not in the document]
vitalybu: You can probably use a regular expression in `checkstyle` to validate that there's a number in this place instead of a hard-coded value.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBase.java
mstrelex: https://user-images.githubusercontent.com/5516364/124916289-1e48dc00-dfb8-11eb-8b8a-154cd8f93aa5.mp4

src/main/java/io/testproject/sdk/internal/rest/messages/ReportItemType.java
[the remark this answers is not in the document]
mstrelex: Indeed you are correct, there is such an option :)
https://github.com/testproject-io/java-opensdk/issues/147

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -48,13 +49,18 @@
...
-    private static final Gson GSON = new GsonBuilder().create();
+    protected static final Gson GSON = new GsonBuilder().create();
 
     /**
      * Progress report delay in seconds.
      */
     private static final int PROGRESS_REPORT_DELAY = 3;
 
+    /**
+     * In case of failure during report - attempt maximum 4 times.
```
mstrelex, line 60: This description is strange, why not just: `Maximum attempts to try sending a report to the Agent`?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -173,24 +182,34 @@ void sendReport(final QueueItem item) {
...
             }
-        }
 
-        // Handle unsuccessful response
-        if (response != null && response.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
-            LOG.error("Agent responded with an unexpected status {} to report: [{}]",
-                    response.getStatusLine().getStatusCode(), item.getReport());
-        }
+            // Handle unsuccessful response
+            if (response != null && response.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
```
mstrelex, line 201: It would be better checking all 2xx codes:
`response.getStatusInfo().getFamily().equals(Response.Status.Family.SUCCESSFUL)`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -173,24 +182,34 @@ void sendReport(final QueueItem item) {
...
+                LOG.error("Agent responded with an unexpected status {} to report: [{}]",
+                        response.getStatusLine().getStatusCode(), item.getReport());
+                reportAttemptsCount--;
+                if (reportAttemptsCount == 0) {
+                    LOG.error("Failed to send reports to the agent.");
```
mstrelex, line 206: Would be more informative: 

`LOG.error("All attempts ({}) sending report [{}] failed", MAX_REPORT_FAILURE_ATTEMPTS , item.getReport(), e);`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -173,24 +182,34 @@ void sendReport(final QueueItem item) {
...
+                    throw new AgentConnectException("Failed to send reports to the agent.");
```
mstrelex, line 207: This is wrong, as the Agent did respond with some code. \
Maybe you want to add a new exception type? Or use `javax.ws.rs.core.Response.WebApplicationException`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -65,63 +70,86 @@
...
+        // Try to get maximum report batch size from env variable.
+        this.maxBatchSize = (System.getenv(TP_MAX_BATCH_SIZE) != null)
+                ? Integer.parseInt(System.getenv(TP_MAX_BATCH_SIZE)) : MAX_REPORTS_BATCH_SIZE;
+    }
 
     /**
      * Handle the report.
      * From version 3.1.0 -> send reports in batches.
      * For lower versions -> Send standalone report
+     * @throws InterruptedException
```
mstrelex, line 83: Missing explanation when and why this exception is thrown.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -65,63 +70,86 @@
...
-                LOG.error("Agent responded with an unexpected status {} to report: [{}]",
-                        response.getStatusLine().getStatusCode(), batchReports);
-            }
+            // Send the report to the agent.
+            // In case of failure - make 3 more attempts.
+            do {
+                try {
+                    response = this.getHttpClient().execute(httpPost);
+                } catch (IOException e) {
+                    LOG.error("Failed to submit report: [{}]", batchReports, e);
```
mstrelex, line 129: What will this print? Something like `List@ABC123` ?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -65,63 +70,86 @@
...
+                } finally {
+                    // Consume response to release the resources
+                    if (response != null) {
+                        EntityUtils.consumeQuietly(response.getEntity());
+                    }
+                    batchReports.clear();
+                }
 
+                // Handle unsuccessful response
+                if (response != null && response.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
```
mstrelex, line 140: Same comments as in the previous commit regarding 2xx status and error handling.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -65,63 +70,86 @@
...
-                if (null != item && batchReports.size() < MAX_REPORTS_BATCH_SIZE) {
-                    batchReports.add(item.getReport());
-                }
-            } while (!getQueue().isEmpty() && batchReports.size() < MAX_REPORTS_BATCH_SIZE);
 
+        while (!getQueue().isEmpty() && batchReports.size() < this.maxBatchSize) {
+            // Get the first item in the queue without blocking it.
+            QueueItem item = getQueue().poll();
+
+            if (null != item && item.getReport() != null) {
```
mstrelex, line 98: And if they are null, don't you want to break the loop?

mstrelex (changes requested):

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
mstrelex: ?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -87,14 +102,43 @@ public ReportsQueue(final CloseableHttpClient httpClient, final String sessionId
         this.sessionId = sessionId;
     }
 
+    /**
+     * Access method to get the queue.
+     * @return the reports queue.
+     */
+    public ArrayBlockingQueue<QueueItem> getQueue() {
```
mstrelex, line 109: Why is it public? Anyone can remove items from the queue is accessed, can it be protected?
Also, a getter method standard description syntax is `Getter method for {@link #field}`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -87,14 +102,43 @@ public ReportsQueue(final CloseableHttpClient httpClient, final String sessionId
...
+        return queue;
+    }
+
+    /**
+     * Access method to get the httpClient.
```
mstrelex, line 114: Same as above.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -72,9 +87,9 @@
     private Future<?> progressFuture;
 
     /**
-     * Progress report delay in seconds.
+     *  In case of 4 failures to send reports to the agent.
      */
-    private static final int PROGRESS_REPORT_DELAY = 3;
+    private boolean brokenReports = false;
```
mstrelex, line 92: Can you think of a better name?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -72,9 +87,9 @@
     private Future<?> progressFuture;
 
     /**
-     * Progress report delay in seconds.
+     *  In case of 4 failures to send reports to the agent.
```
mstrelex, line 90: A proper description would be - `A flag that is raised when all attempts submitting a report fail.`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -136,26 +182,37 @@ private void sendReport(final QueueItem item) {
...
+            if (response != null && response.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
+                LOG.error("Agent responded with an unexpected status {} to report: [{}]",
+                        response.getStatusLine().getStatusCode(), item.getReport());
+                reportAttemptsCount--;
+                if (reportAttemptsCount == 0) {
+                    LOG.error("Failed to send reports to the agent.");
+                    throw new AgentConnectException("Failed to send reports to the agent.");
+                }
+
+                LOG.error("Attempt to send report again to the Agent. {} more attempts are left.", reportAttemptsCount);
```
mstrelex, line 210: Please change to warning log and the wording to:

`Failed to send a report to the Agent, {} attempts remaining...`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
mstrelex: Will be addressed in #147

mstrelex (changes requested):

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
oriddd: I'm not sure. 
This check is to make sure the report isn't null - which we are not expecting to. 
We're except to get null only once adding the null item which was insert in order to release the queue.
In that case - the while condition is checking if the queue is empty - and the loop won't continue.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
oriddd: I can't use that since javax.ws.rs.core.* is not included in the project.
Therefor I've created new Exception type for report failure.

src/main/java/io/testproject/sdk/internal/exceptions/FailedReportException.java
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.testproject.sdk.internal.exceptions;
+
+/**
+ * Exception object thrown when SDK is failed to send reports to the agent.
```
mstrelex, line 21: ---> `This exception is thrown when SDK fails to send a report to the agent.`

src/main/java/io/testproject/sdk/internal/exceptions/FailedReportException.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Default serial version ID.
+     */
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Creates a new instance of the class with a default message.
+     */
+    public FailedReportException() {
+        this("Failed to send report to the Agent");
```
mstrelex, line 33: ---> `Failed sending a report to the Agent`

mstrelex (changes requested):

src/main/java/io/testproject/sdk/internal/rest/AgentClient.java
```diff
@@ -210,6 +210,12 @@
      */
     private SessionResponse agentResponse;
 
+    /**
+     * Minimum Agent version that supports reporting at batches.
```
mstrelex, line 214: `Minimum Agent version that supports reporting at batches.` --> `Minimum Agent version that supports batch reporting.`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -186,9 +254,21 @@ public void stop() {
...
 
     /**
-     * Internal class to keep the HTTP request and contained report together in the queue.
+     * Helper method to check if CloseableHttpResponse status code is success.
+     * @param response
+     * @return true if HTTP status is in range of [200-299].
+     * Otherwise return false.
      */
-    private static class QueueItem {
+    protected boolean isHTTPResponseSuccess(final CloseableHttpResponse response) {
```
mstrelex, line 262: Please do not reinvent the wheel, I already wrote to you how to do it the right way - `getStatusInfo().getFamily() != Family.SUCCESSFUL`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -44,7 +49,22 @@
...
+
+    /**
+     * Maximum attempts to try sending a report to the Agent.
+     */
+    protected static final int MAX_REPORT_FAILURE_ATTEMPTS = 4;
+
+    /**
+     * Last http status which defined as success status.
+     */
+    protected static final int LAST_SUCCESS_HTTP_STATUS = 299;
```
mstrelex, line 67: Irellevant, use `getStatusInfo().getFamily() == Family.SUCCESSFUL`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -87,14 +107,43 @@ public ReportsQueue(final CloseableHttpClient httpClient, final String sessionId
...
      */
     void submit(final HttpEntityEnclosingRequestBase request, final Report report) {
-        this.queue.add(new QueueItem(request, report));
+        if (!this.multipleFailuresToReport) {
+            this.queue.add(new QueueItem(request, report));
+        }
+    }
+
+    /**
+     * Handle the report.
```
mstrelex, line 139: You are stating the obvious, this is not a summary, the same can be inferred from the method name.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -136,26 +187,43 @@ private void sendReport(final QueueItem item) {
...
+                // Consume response to release the resources
+                if (response != null) {
+                    EntityUtils.consumeQuietly(response.getEntity());
+                }
             }
-        }
 
-        // Handle unsuccessful response
-        if (response != null && response.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
+            if (isHTTPResponseSuccess(response)) {
```
mstrelex, line 207: `getStatusInfo().getFamily() == Family.SUCCESSFUL`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -136,26 +187,43 @@ private void sendReport(final QueueItem item) {
...
+            // Handle unsuccessful response
             LOG.error("Agent responded with an unexpected status {} to report: [{}]",
                     response.getStatusLine().getStatusCode(), item.getReport());
+            reportAttemptsCount--;
+            LOG.error("Failed to send a report to the Agent, {} attempts remaining...", reportAttemptsCount);
+        } while (!isHTTPResponseSuccess(response) && reportAttemptsCount > 0);
+
+        // In case all attempts to send the report are failed.
+        if (reportAttemptsCount == 0) {
+            LOG.error("All {} attempts to send report {} are failed", MAX_REPORT_FAILURE_ATTEMPTS, item.getReport());
```
mstrelex, line 220: `All {} attempts to send report {} are failed` --> `All {} attempts to send report {} have failed`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Default batch report size is maximum 10 reports.
+     */
+    private static final int MAX_REPORTS_BATCH_SIZE = 10;
+
+    /**
+     * Constant for environment variable name that may store the max batch size.
+     */
+    private static final String TP_MAX_BATCH_SIZE = "TP_MAX_BATCH_SIZE";
```
mstrelex, line 55: `TP_MAX_BATCH_SIZE` ---> `TP_MAX_BATCH_SIZE_VARIABLE_NAME`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,156 @@
...
+
+                if (isHTTPResponseSuccess(response)) {
+                    break;
+                }
+
+                // Handle unsuccessful response
+                LOG.error("Agent responded with an unexpected status {} to report: [{}]",
+                       response.getStatusLine().getStatusCode(), batchReports);
+                reportAttemptsCount--;
+                LOG.error("Attempt to send report again to the Agent. {} more attempts are left.", reportAttemptsCount);
```
mstrelex, line 145: Please change to INFO:

```
LOG.info("Retrying ({}) sending reports batch to the Agent...", MAX_REPORT_FAILURE_ATTEMPTS 
 - reportAttemptsCount);
 ```

File: src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java [hunk 54]
```diff
@@ -0,0 +1,156 @@
...
+            } while (!isHTTPResponseSuccess(response) && reportAttemptsCount > 0);
+
+            // In case all attempts to send the report are failed.
+            if (reportAttemptsCount == 0) {
+                LOG.error("All {} attempts to send report {} are failed", MAX_REPORT_FAILURE_ATTEMPTS, json);
```
mstrelex, line 150: You should not print the whole JSON in the error, what is it good for?

mstrelex (changes requested):

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -136,23 +160,43 @@ private void sendReport(final QueueItem item) {
...
+                }
+
+                // Handle unsuccessful response
+                LOG.warn("Agent responded with an unexpected status {} to report: [{}]",
+                        response.getStatusLine().getStatusCode(), item.getReport());
+                reportAttemptsCount--;
+                LOG.info("Failed to send a report to the Agent, {} attempts remaining...", reportAttemptsCount);
             }
-        }
+        } while (response != null && Response.Status.Family.familyOf(response.getStatusLine().getStatusCode())
```
mstrelex, line 208: Shouldn't it be `or`? While the response is null, or it's unsuccessful and there are more attempts left?

mstrelex (changes requested):

mstrelex:
Please adjust commit messages:

0f997ed9210102c583ac6eb79191103b12ccd9f4 --> `feat: Adding new route to support reports in batches.`
bcedcb9ae2cf560870d818fd153aff01f7c0ef12 --> `feat: Change QueueItem class to internal in order to use in additional classes.`
cc9e35d77c442ebf81f04d31517759a21222c9ac --> `feat: Prepare ReportsQueue class to be extended by adding new method … which can be overridden.`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * The default batch report size is a maximum of 10 reports.
+     */
+    private static final int MAX_REPORTS_BATCH_SIZE = 10;
+
+    /**
+     * Constant for environment variable name that may store the max batch size.
+     */
+    private static final String TP_MAX_BATCH_SIZE_VARIABLE_NAME = "TP_MAX_BATCH_SIZE";
```
vitalybu, line 56: `TP_MAX_BATCH_SIZE` --> `TP_MAX_REPORTS_BATCH_SIZE`

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,167 @@
...
+     * @param httpClient HTTP client ot use for communicating with the Agent.
+     * @param sessionId  Driver session ID.
+     * @param remoteAddress Reports remote address.
+     */
+    ReportsQueueBatch(final CloseableHttpClient httpClient, final String sessionId, final URL remoteAddress) {
+        super(httpClient, sessionId);
+        this.remoteAddress = remoteAddress;
+
+        // Try to get maximum report batch size from env variable.
+        this.maxBatchSize = (System.getenv(TP_MAX_BATCH_SIZE_VARIABLE_NAME) != null)
```
vitalybu, line 75: Should surround this with a `try-catch` and use the default value on failure.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
oriddd: You're right. 
I've added this null check due to spotbug failures. 
Will change and commit as fixup.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,167 @@
...
+     * @throws FailedReportException in case of 4 failures to send reports to the agent
+     */
+    @Override
+    void handleReport() throws InterruptedException, FailedReportException {
+        // Linked list to store the reports batch before sending them.
+        // Those reports will be extract from the queue.
+        List<Report> batchReports = new LinkedList<>();
+
+        // Extract and remove up to 10 items or till queue is empty from queue - without blocking it.
+
```
vitalybu, line 93: Remove empty line

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,167 @@
...
+        while (!getQueue().isEmpty() && batchReports.size() < this.maxBatchSize) {
+            // Get the first item in the queue without blocking it.
+            QueueItem item = getQueue().poll();
+
+            if (null != item && item.getReport() != null) {
+                batchReports.add(item.getReport());
+            }
+        }
+
+        if (!batchReports.isEmpty()) {
```
vitalybu, line 103: To reduce code indentation it's probably better to return from this method if `batchReports.isEmpty() == true`...

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,167 @@
...
+                    response = this.getHttpClient().execute(httpPost);
+                } catch (IOException e) {
+                    LOG.error("Failed to submit {} reports.", numberOfReports, e);
+                    return;
+                } finally {
+                    // Consume response to release the resources
+                    if (response != null) {
+                        EntityUtils.consumeQuietly(response.getEntity());
+                    }
+                    batchReports.clear();
```
vitalybu, line 137: Why?

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,167 @@
...
+                    // Handle unsuccessful response
+                    LOG.warn("Agent responded with an unexpected status {} to send {} reports",
+                            response.getStatusLine().getStatusCode(), numberOfReports);
+
+
+                    reportAttemptsCount--;
+                    LOG.info("Attempt to send {} reports again to the Agent. {} more attempts are left.",
+                            numberOfReports, reportAttemptsCount);
+                }
+            } while (response != null && Response.Status.Family.familyOf(response.getStatusLine().getStatusCode())
```
vitalybu, line 155: Should probably be ORs as @mstrelex mentioned before

vitalybu (changes requested):

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -103,13 +124,15 @@ void submit(final HttpEntityEnclosingRequestBase request, final Report report) {
...
+        while (this.running || !this.queue.isEmpty()) {
             try {
-                sendReport(queue.take());
+                handleReport();
             } catch (InterruptedException e) {
                 LOG.error("Reports queue was interrupted");
                 break;
+            } catch (FailedReportException e) {
+                this.stopReports = true;
+                LOG.warn("Reports are disabled due to multiple failed attempts of sending reports to the agent.");
```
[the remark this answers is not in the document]
oriddd: Before sending the reports - we are printing the exception message as an error. 
This warning is to let the user know that his reports are disabled from now on.
This warning was added to prevent bugs open due to missing reports in case of multiple failing attempts to report to the agent.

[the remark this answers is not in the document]
vitalybu: 👍

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -44,7 +44,17 @@
     /**
      * An instance of the Google JSON serializer to serialize and deserialize objects.
      */
-    private static final Gson GSON = new GsonBuilder().create();
+    protected static final Gson GSON = new GsonBuilder().create();
+
+    /**
+     * Progress report delay in seconds.
```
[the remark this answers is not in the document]
mstrelex: We don't delay the report, we have a delay between progress logs.
When the process using the SDK finishes, it logs how many reports are left to be reported.
To avoid noise in the log and too frequent progress reporting, we have a delay between the **progress** reports.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
```diff
@@ -0,0 +1,130 @@
...
+        this.remoteAddress = remoteAddress;
+
+        // Try to get maximum report batch size from env variable.
+        int maxBatchSizeValueToSet;
+        try {
+            String tpMaxBatchSizeEnvVarValue = System.getenv(TP_MAX_BATCH_SIZE_VARIABLE_NAME);
+            maxBatchSizeValueToSet = (tpMaxBatchSizeEnvVarValue != null)
+                    ? Integer.parseInt(tpMaxBatchSizeEnvVarValue) : MAX_REPORTS_BATCH_SIZE;
+        } catch (Exception e) {
+            LOG.info("Failed to retrieve the value of environment variable {}. "
```
mstrelex, line 77: Should be a warning and the last argument should be the exception.
What can go wrong here anyway? Failure to convert string to an integer?

mstrelex (changes requested):

src/main/java/io/testproject/sdk/internal/rest/ReportsQueueBatch.java
oriddd: You're right. 
The failure might be in:
1) SecurityException -  if a security manager exists and its checkPermission method doesn't allow access to the process environment (not expected to be that case).
2) Convert the string to an integer.

src/main/java/io/testproject/sdk/internal/rest/ReportsQueue.java
```diff
@@ -122,38 +176,47 @@ public void run() {
...
+     * @param httpPost For lower versions than 3.1.0 -> HTTP request retrieved from the queue.
+     *                 For versions 3.1.0 and greater -> Build reports batch HTTP request.
+     * @throws FailedReportException in case of 4 failures to send reports to the agent
      */
-    private void sendReport(final QueueItem item) {
-        if (item.getRequest() == null && item.getReport() == null) {
-            if (this.running) {
-                // There nulls are not OK, something went wrong preparing the report/request.
-                LOG.error("Empty report and request were submitted to the queue!");
+    void sendReport(final HttpPost httpPost) throws FailedReportException {
```
vitalybu, line 183: Does this method should really be `public`?
If it's used internally, then it should be `protected`, no?

oriddd: Right. 
Please correct me if I'm wrong, the default (without any modifier) is package-protect which should be more secure, no?

vitalybu: You're right, forgot this is the default behavior in Java.
Let's leave it as is.

mstrelex (approved):

vitalybu (approved):

mstrelex (approved):

Issue #151, opened 2021-08-10
Setting tp:guid capability on Internet Explorer driver causes an exception


Artem-Kuz:
When setting tp:guid capability on Internet Explorer driver, it rejects it and throws an exception:
```
io.testproject.sdk.internal.exceptions.AgentConnectException: Agent responded with status 500: [No matching capability sets found.
Unable to match capability set 0: tp:guid is an unknown extension capability for IE
```

Artem-Kuz:
This will be fixed in the agent side and released in 3.3.0 version

PR #152, opened 2021-08-10
fix: avoid setting tp:guid capability if test is executed on IE

Artem-Kuz:
relates to #151 
Signed-off-by: Artem Kuznetsov <artem.kuznetsov@testproject.io>


mstrelex:
@Artem-Kuz 

Can you please elaborate on this PR?
If you will remove the `TP_GUID` how the AgentCLient instance will be tracked?

IMHO, if IE doesn't like a capability, it should be removed in the Agent, that initiates the driver and not the SDK who reconstructs it.

s-glatshtein:
@Artem-Kuz did we not agree that this should be done in agent side and not as part of the SDK?

Artem-Kuz:
Agreed

PR #172, opened 2022-01-27
Split workflow to dedicated workflows that are triggering a reusable workflow

oriddd:
Split workflow to dedicated workflows that are triggering a reusable workflow. 
build_release.yml is the workflow for release. 
build_snapshot.yml is the workflow for every push.
build.yml is the reusable workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,13 +1,10 @@
 # This workflow builds the SDK using Java 11
 name: Build
 
+# This is a reusable workflow.
 on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
-  release:
-    types: [published]
+  workflow_call:
+
 jobs:
   # Build and test the SDK
   build-test-publish:
@@ -86,35 +83,15 @@ jobs:
       with:
         name: reports
         path: build/reports
-    - name: Define uberJAR name
+    - name: Define uberJAR file name
       if: startsWith(github.ref, 'refs/tags/v')
       run: |
-        export TP_SDK_VARIANT="RELEASE"
-        echo "TP_SDK_VARIANT=RELEASE" >> $GITHUB_ENV
-        export TP_SDK_JAR_NAME="testproject-java-sdk"
-        export TP_SDK_UBER_JAR_EXT="with-dependencies.jar"
-        echo "TP_SDK_JAR_NAME=testproject-java-sdk" >> $GITHUB_ENV
-        echo "TP_SDK_UBER_JAR_EXT=with-dependencies.jar" >> $GITHUB_ENV
-        echo "TP_SDK_JAR_EXT=.jar" >> $GITHUB_ENV
-    - name: Upload uberJAR to release
+        export TP_SDK_JAR_NAME="testproject-java-sdk.jar"
+        echo "TP_SDK_JAR_NAME=testproject-java-sdk.jar" >> $GITHUB_ENV
+        mkdir -p build/libs/upload && cp build/libs/*with-dependencies.jar build/libs/upload/"${TP_SDK_JAR_NAME}"
+    - name: Upload uberJAR workflow artifact
       if: startsWith(github.ref, 'refs/tags/v')
-      uses: svenstaro/upload-release-action@v2
+      uses: actions/upload-artifact@v2
       with:
-        repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: build/libs/${{ env.TP_SDK_JAR_NAME }}-${{ env.TP_SDK_VERSION }}-${{ env.TP_SDK_VARIANT}}-${{ env.TP_SDK_UBER_JAR_EXT }}
-        asset_name: ${{ env.TP_SDK_JAR_NAME }}${{ env.TP_SDK_JAR_EXT }}
-        tag: ${{ github.ref }}
-        overwrite: true
-    - name: Create uberJAR hash file
-      if: startsWith(github.ref, 'refs/tags/v')
-      run: |
-        md5sum build/libs/${{ env.TP_SDK_JAR_NAME }}-${{ env.TP_SDK_VERSION }}-${{ env.TP_SDK_VARIANT}}-${{ env.TP_SDK_UBER_JAR_EXT }} > build/libs/jar-hash.md
-    - name: Upload hash artifact
-      if: startsWith(github.ref, 'refs/tags/v')
-      uses: svenstaro/upload-release-action@v2
-      with:
-        repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: build/libs/jar-hash.md
-        asset_name: ${{ env.TP_SDK_JAR_NAME }}.md5
-        tag: ${{ github.ref }}
-        overwrite: true
+        name: ${{ env.TP_SDK_JAR_NAME }}
+        path: build/libs/upload/${{ env.TP_SDK_JAR_NAME }}
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -1,0 +1,44 @@
+# This workflow is triggering the reusable build workflow.
+# It's also extend the logic to upload uberJAR, and it's hash as release artifacts.
+
+name: BuildRelease
+
+on:
+  release:
+    types: [created]
+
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -144,6 +144,25 @@ jar {
...
+    }
+
+    archiveFileName = project.name + "-" + project.version + "-FULL.jar"
+    from configurations.compileClasspath.collect {
+        it.isDirectory() ? it : zipTree(it)
+    }
+    with jar
+}
+
+// Add the fullJar task to the build lifecycle.
```
mstrelex, line 163: Rename `fullJar`

.github/workflows/build.yml
```diff
@@ -86,3 +86,41 @@ jobs:
       with:
         name: reports
         path: build/reports
+    - name: Define FULL jar name
```
mstrelex, line 89: This is a pure CI commit, it's not `build:`, it's `ci:`
Also, this should not be part of the build workflow, but a separate workflow that is triggered when the SDK is published.

Please adjust the commit messages and the description as well.
Rename Full Jar to uberJAR

.github/workflows/build.yml
```diff
@@ -86,3 +86,41 @@ jobs:
...
+        if [ -z "$SDK_BUILD_ARGS" ]
+        then
+        export TP_SDK_VARIANT="SNAPSHOT"
+        echo "TP_SDK_VARIANT=SNAPSHOT" >> $GITHUB_ENV
+        else
+        export TP_SDK_VARIANT="RELEASE"
+        echo "TP_SDK_VARIANT=RELEASE" >> $GITHUB_ENV
+        fi
+        export TP_SDK_JAR_NAME="testproject-java-sdk"
+        export TP_SDK_FULL_JAR_EXT="FULL.jar"
```
mstrelex, line 101: No need for the `FULL` suffix.

.github/workflows/build.yml
```diff
@@ -86,3 +86,41 @@ jobs:
       with:
         name: reports
         path: build/reports
+    - name: Define FULL jar name
+      if: startsWith(github.ref, 'refs/tags/v')
+      run: |
+        if [ -z "$SDK_BUILD_ARGS" ]
+        then
+        export TP_SDK_VARIANT="SNAPSHOT"
```
mstrelex, line 94: No need to publish uberJAR for SNAPSHOT releases.

.github/workflows/build.yml
```diff
@@ -86,3 +86,41 @@ jobs:
       with:
         name: reports
         path: build/reports
+    - name: Define FULL jar name
+      if: startsWith(github.ref, 'refs/tags/v')
```
mstrelex, line 90: Can be removed if the whole flow runs on published releases only.

.github/workflows/build.yml
```diff
@@ -86,3 +86,41 @@ jobs:
...
+        else
+        export TP_SDK_VARIANT="RELEASE"
+        echo "TP_SDK_VARIANT=RELEASE" >> $GITHUB_ENV
+        fi
+        export TP_SDK_JAR_NAME="testproject-java-sdk"
+        export TP_SDK_FULL_JAR_EXT="FULL.jar"
+        echo "TP_SDK_JAR_NAME=testproject-java-sdk" >> $GITHUB_ENV
+        echo "TP_SDK_FULL_JAR_EXT=FULL.jar" >> $GITHUB_ENV
+        echo "TP_SDK_JAR_EXT=.jar" >> $GITHUB_ENV
+    - name: Upload FULL jar to release
```
mstrelex, line 105: Rename, please.

.github/workflows/build.yml
```diff
@@ -86,3 +86,41 @@ jobs:
...
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        file: build/libs/${{ env.TP_SDK_JAR_NAME }}-${{ env.TP_SDK_VERSION }}-${{ env.TP_SDK_VARIANT}}-${{ env.TP_SDK_FULL_JAR_EXT }}
+        asset_name: ${{ env.TP_SDK_JAR_NAME }}${{ env.TP_SDK_JAR_EXT }}
+        tag: ${{ github.ref }}
+        overwrite: true
+    - name: Create jar's hash file
+      if: startsWith(github.ref, 'refs/tags/v')
+      run: |
+        md5sum build/libs/${{ env.TP_SDK_JAR_NAME }}-${{ env.TP_SDK_VERSION }}-${{ env.TP_SDK_VARIANT}}-${{ env.TP_SDK_FULL_JAR_EXT }} > build/libs/jar-hash.md
```
mstrelex, line 117: Why save the hash to file and not variable?

mstrelex (changes requested):

build.gradle
```diff
@@ -144,6 +144,25 @@ jar {
     }
```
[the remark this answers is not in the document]
s-glatshtein: @mstrelex there is such a thing and it applies to gradle changes only
@oriddd that means your commit fixing the github build script should be `ci:` and not `build`

build.gradle
```diff
@@ -144,6 +144,25 @@ jar {
...
+task fullJar (type: Jar) {
+    mustRunAfter(jar, javadocJar, sourcesJar)
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+
+    manifest {
+        attributes 'Compile-Title': project.name,
+                'Compile-Version': project.version
+    }
+
+    archiveFileName = project.name + "-" + project.version + "-FULL.jar"
```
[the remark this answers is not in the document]
oriddd: This file (without the suffix) already exists (since the jar task is creating it). 
I'll replace the "FULL" suffix with "UBER"

.github/workflows/build.yml
oriddd: Please see my comment: 
https://github.com/testproject-io/java-opensdk/pull/172/commits#r793807676

.github/workflows/build.yml
oriddd: According to the architecture, we need to compare the SDK uber jars according to the md5 hash. 
This md5 hash is saved as an artifact to avoid downloading the whole jar and calculating the hash each time.

.github/workflows/build.yml
oriddd: @mstrelex 
Please review the example I've created in the private repository: 
https://github.com/oriddd/java-opensdk/tree/master/.github/workflows 

- The **build.yml** is a reusable workflow.

- The **build_snapshot.yml** is getting triggered by every commit and running the build.yml. 

- The **build_release.yml** is getting triggered only once a release is created. 
It's running the buid.yml and adding another logic after that. 

The main issue is: 
Those 2 jobs are running on separate runner machines and the second job (that needs to upload the uberJAR file) doesn't have this jar and must to re-build it.  
Do you have any idea how can we solve that problem without re-building (running grade build again)?

oriddd:
@mstrelex @s-glatshtein  
This is currently blocked by https://github.community/t/ref-head-in-reusable-workflows/203690/67 
Will create new PR with the old approach that has 1 workflow file with actions that are running only for release. 
Marking this PR as WIP for future changes once the bug will be fixed.

build.gradle
```diff
@@ -144,6 +144,25 @@ jar {
...
+task uberJar (type: Jar) {
+    mustRunAfter(jar, javadocJar, sourcesJar)
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+
+    manifest {
+        attributes 'Compile-Title': project.name,
+                'Compile-Version': project.version
+    }
+
+    archiveFileName = project.name + "-" + project.version + "-UBER.jar"
```
mstrelex, line 156: Why do we need the suffix? `-UBER.jar`

mstrelex (changes requested):

build.gradle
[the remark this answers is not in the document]
mstrelex: I understand now. The Standard IMHO is something lower case, e.g. `jar-with-dependencies`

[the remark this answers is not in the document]
oriddd: It's changed as part of: https://github.com/testproject-io/java-opensdk/pull/173 
The changes on build.gradle file were removed from this PR :)

